PR #5, opened 2016-12-31
Enable continuous integration with TravisCI

ianmkenney:
[Alchemlyb TravisCI Panel](https://travis-ci.org/alchemistry/alchemlyb)

Python Versions:
* 2.7

Branches:
* master
* develop (not a major branch yet)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[run]
+branch = True
+omit =
+    # omit all tests
+    src/alchemlyb/tests/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: python
+sudo: false
+
+python:
+   - "2.7"
+
+branches:
+    only:
+        - master
+        - develop
+
+install:
+    - pip install --upgrade pip setuptools wheel
+    - pip install codecov
+    - pip install pytest-cov
+    - pip install pytest-pep8
+    - pip install --only-binary=numpy numpy # Otherwise this would take ages
+    - pip install -e .
+
+script:
+    - py.test --cov alchemlyb alchemlyb/src/alchemlyb/tests
+
+after_success:
+    - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 
 python:
    - "2.7"
+   - "3.5"
 
 branches:
     only:
@@ -11,6 +12,7 @@ branches:
 
 install:
     - pip install --upgrade pip setuptools wheel
+    - pip install six
     - pip install codecov
     - pip install pytest-cov
     - pip install pytest-pep8
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
     - pip install -e .
 
 script:
-    - py.test --cov alchemlyb alchemlyb/src/alchemlyb/tests
+    - py.test --cov alchemlyb src/alchemlyb/tests
 
 after_success:
     - codecov
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 alchemlyb: data munging and analysis library for alchemical free energy calculations
 ====================================================================================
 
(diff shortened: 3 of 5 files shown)


ianmkenney:
@dotsdl It looks like pymbar doesn't support python 3 syntax
```
Collecting pymbar (from alchemlyb==0.1.0.dev0)

  Downloading pymbar-2.1.0-beta.tar.gz (50kB)

    100% |████████████████████████████████| 51kB 6.4MB/s 

    Complete output from command python setup.py egg_info:

    Traceback (most recent call last):

      File "<string>", line 1, in <module>

      File "/tmp/pip-build-2j9i1paj/pymbar/setup.py", line 136, in <module>

        main()

      File "/tmp/pip-build-2j9i1paj/pymbar/setup.py", line 132, in main

        setupKeywords = buildKeywordDictionary()

      File "/tmp/pip-build-2j9i1paj/pymbar/setup.py", line 121, in buildKeywordDictionary

        for key in sorted( setupKeywords.iterkeys() ):

    AttributeError: 'dict' object has no attribute 'iterkeys'
```

I assume this means that we can only test python 2

ianmkenney:
@dotsdl Please review.

davidlmobley:
As far as I know, `pymbar` should support Python 3; at least, @jchodera is trying to make everything python3 compatible these days. Perhaps he can advise.

davidlmobley:
Also, I think @mrshirts should probably be able to comment since he's one of the `pymbar` devs as well. I think at this point it's critical we move to python3 as python2 support is going away starting this year in many major packages, IIRC.

davidlmobley:
It's in the [https://github.com/choderalab/pymbar/blob/master/.travis.yml](`.travis.yml`) for pymbar, so it looks like they're trying to support it at least. I suggest an issue on the issue tracker.

mrshirts:
I believe it's just the setup that doesn't support python 3.  I think John and Levi are working on this now?

jchodera:
What is the precise problem? Can you paste an example of what you are trying to do that doesn't work?

jchodera:
I know of no python 3 issues with pymbar.

jchodera:
Ah, I just saw https://github.com/alchemistry/alchemlyb/pull/5#issuecomment-269848687.

That's `pymbar-2.1.0-beta`, which is not python 3.x compatible. You'll have to use the `pymbar 3` release, which was currently only available via github or `conda` installs.

I've pushed the `pymbar-3.0.0.dev0` package to pypi, since I think that's what you're using:
https://pypi.python.org/pypi/pymbar/3.0.0.dev0

We're in the process of cleaning up the documentation and have a full 3.0.0 release scheduled for Friday, where we'll also update the license to MIT.

ianmkenney:
I will add python 3.5 to testing since we expect it to work in the near future. Thanks!

jchodera:
Looks like `six` is imported in `setup.py` but wasn't added to `requirements`. I've created a PR to fix: https://github.com/choderalab/pymbar/pull/251

jchodera:
We're still hoping to update the release on Friday, so I would suggest you simply `pip install six` before you install `pymbar` at this point to work around this issue in the meantime.

ianmkenney:
See latest build for 2.7 for compilation error.

jchodera:
That's very odd. Trying to replicate locally.

ianmkenney:
I can try it in 6-7 hours once I'm back from my family New Years event

jchodera:
I've disabled the C extensions and pushed a new `3.0dev3` package to pypi. You may need to use the `--pre` flag for `pip` to pick this version up.

ianmkenney:
@jchodera reran the build and it worked. Thanks for the assistance!

jchodera:
Great! Sorry for the hassle here---you uncovered a lot of issues that we hadn't realized existed.

dotsdl:
Awesome, thanks all for this! Merging.

Issue #16, opened 2017-02-18
Remove GromacsWrapper dependency from alchemlyb.parsers.gmx


dotsdl:
We currently use the `gromacs.fileformats.XVG` class for parsing out data from XVG files in the `alchemlyb.parsers.gmx` module. This works well, but it means [GromacsWrapper](https://github.com/Becksteinlab/GromacsWrapper) is a dependency and it is not yet working under [Python 3](https://github.com/Becksteinlab/GromacsWrapper/issues/44). This holds back `alchemlyb` from full Python 3 support.

I recommend implementing a helper function in `alchemlyb.parsers.gmx` that parses out the raw tabular information of an XVG file similarly to how the GromacsWrapper XVG class does this in its `XVG.to_df()` method. We cannot outright lift the code from GromacsWrapper due to its licensing, but we can reimplement the (minimal) functionality we need here.

orbeckst:
I think this can be closed after the merge of #19.

PR #19, opened 2017-02-25
standalone simple XVG parser

orbeckst:
Bare-bones XVG parser that will make alchemlyb independent from GromacsWrapper #16 (which is GPLv2 licensed and not compatible with BSD-3 of alchemlyb). I originally wrote all the code and explicitly relicense it under BSD-3.


src/alchemlyb/parsing/util.py
```diff
@@ -33,9 +33,9 @@ def anyopen(filename, mode='r'):
...
 
     if ext in extensions:
-       opener = extensions[ext] 
+       opener = extensions[ext]
 
     else:
         opener = open
 
-    return opener(filename, 'r')
+    return opener(filename, mode)
```
dotsdl: Thanks for the catch!

setup.py
```diff
@@ -1,5 +1,5 @@
 #! /usr/bin/python
-"""Setuptools-based setup script for datreant.
+"""Setuptools-based setup script for alchemlyb.
```
dotsdl: Nice catch; thanks! (obvious reuse of some structure on my part)

PR #32, opened 2017-10-30
add amber TI parser and improve the subsampling code

### LINKED ISSUE #10: AMBER parsers
In line with the overall [API proposal](http://alchemlyb.readthedocs.io/en/latest/api_proposal.html), we want to have parsers for each of the major MD engines, and eventually have coverage for all of those in use. Since there are essentially two types of estimators (TI and FEP), each packages needs a parser for:
1. ~Extracting reduced potentials `u_nk` from output files (for FEP).~ (EDIT: This objective is postponed, emphasis right now  on TI. Open issue for `u_nk` when needed. — @orbeckst 2017-11-03)
2. Extracting derivatives `DHdl` from output files (for TI).

This issue is the nexus for discussion for such parsers for the **AMBER** package. If you have existing parsing code for this package, comment below and we can begin adapting it into the parsers outlined above in a PR.

shuail:
Fixes #10 

1. add the amber TI parser to get the dhdl
2. Inside the statistical_inefficiency function under subsampling.py, switch to use the subsampleCorrelatedData in pymbar to generate the subsampled data points, the original code use the pandas slicing method which required to round the statistical inefficiency to an integer.

### DIFF
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -3,12 +3,16 @@ Most of the file parsing part are inheriting from alchemical-analysis
 Change the final format to pandas to be consistent with the alchemlyb format
 """
 
-import pandas as pd
-import re
-import numpy as np
 import os
+import re
+import pandas as pd
+import numpy as np
+import logging 
 
-def convert_to_pandas(file_datum, ):
+logger = logging.getLogger("alchemlyb.parsers.Amber")
+
+def convert_to_pandas(file_datum):
+    """Convert the data structure from numpy to pandas format"""
     data_dic = {}
     data_dic["dHdl"] = []
     data_dic["lambdas"] = []
@@ -68,9 +72,10 @@ class SectionParser(object):
         try:
             self.fileh = open_it(self.filename, 'rb')
             self.filesize = os.stat(self.filename).st_size
-        except IOError:
-            raise SystemExit('ERROR: cannot open file %s' % filename)
+        except Exception as ex:
+            logging.exception("ERROR: cannot open file %s" % filename)
         self.lineno = 0
+
     def skip_lines(self, nlines):
         """Skip a given number of files."""
         lineno = 0
@@ -79,6 +84,7 @@ class SectionParser(object):
             if lineno > nlines:
                 return line
         return None
+
     def skip_after(self, pattern):
         """Skip until after a line that matches a regex pattern."""
         for line in self:
@@ -86,6 +92,7 @@ class SectionParser(object):
             if match:
                 break
         return self.fileh.tell() != self.filesize
+
     def extract_section(self, start, end, fields, limit=None, extra='',
                         debug=False):
         """
@@ -122,8 +129,10 @@ class SectionParser(object):
             else:                       # section may be incomplete
                 result.append(None)
         return result
+
     def __iter__(self):
         return self
+
     def next(self):
         """Read next line of the filehandle and check for EOF."""
         self.lineno += 1
@@ -132,19 +141,20 @@ class SectionParser(object):
             raise StopIteration
         # NOTE: can't mix next() with seek()
         return self.fileh.readline()
+
     def close(self):
         """Close the filehandle."""
         self.fileh.close()
+
     def __enter__(self):
         return self
+
     def __exit__(self, typ, value, traceback):
         self.close()
 
 class FEData(object):
     """A simple struct container to collect data from individual files."""
 
-    #__slots__ = ['clambda', 't0', 'dt', 'T', 'gradients',
-    #             'component_gradients', 'mbar_energies']
     __slots__ = ['clambda', 't0', 'dt', 'T', 'gradients',
                  'component_gradients']
 
@@ -155,19 +165,18 @@ class FEData(object):
         self.T = -1.0
         self.gradients = []
         self.component_gradients = []
-        #self.mbar_energies = []
 
-
-def file_validation(outfile, ):
+def file_validation(outfile):
+    """validate the energy output file """
     invalid = False
     with SectionParser(outfile) as secp:
         line = secp.skip_lines(5) 
         if not line:
-            print('  WARNING: file does not contain any useful data, '
-                      'ignoring file')
+            logging.warning('  WARNING: file does not contain any useful data, '
+                            'ignoring file')
             invalid = True
         if not secp.skip_after('^   2.  CONTROL  DATA  FOR  THE  RUN'):
-            print('  WARNING: no CONTROL DATA found, ignoring file')
+            logging.warning('  WARNING: no CONTROL DATA found, ignoring file')
             invalid = True
         ntpr, = secp.extract_section('^Nature and format of output:', '^$',
                                      ['ntpr'])
@@ -176,85 +185,93 @@ def file_validation(outfile, ):
         T, = secp.extract_section('temperature regulation:', '^$',
                                  ['temp0'])
         if not T:
-            raise SystemExit('ERROR: Non-constant temperature MD not '
-                             'currently supported')
+            logging.error('ERROR: Non-constant temperature MD not '
+                          'currently supported')
             invalid = True
         clambda, = secp.extract_section('^Free energy options:', '^$',
                                         ['clambda'], '^---')
         if clambda is None:
-            print('  WARNING: no free energy section found, ignoring file')
+            logging.warning('  WARNING: no free energy section found, ignoring file')
             invalid = True
 
         if not secp.skip_after('^   3.  ATOMIC '):
-            print('  WARNING: no ATOMIC section found, ignoring file\n')
+            logging.warning('  WARNING: no ATOMIC section found, ignoring file\n')
             invalid = True
 
         t0, = secp.extract_section('^ begin time', '^$', ['coords'])
         if not secp.skip_after('^   4.  RESULTS'):
-            print('  WARNING: no RESULTS section found, ignoring file\n')
+            logging.warning('  WARNING: no RESULTS section found, ignoring file\n')
             invalid = True
     if invalid:
         return False
-    else:
-        file_datum = FEData()
-        file_datum.clambda = clambda
-        file_datum.t0 = t0
-        file_datum.dt = dt
-        file_datum.T = T
-        return file_datum
+    file_datum = FEData()
+    file_datum.clambda = clambda
+    file_datum.t0 = t0
+    file_datum.dt = dt
+    file_datum.T = T
+    return file_datum
 
-def extract_dHdl(outfile, ):
+def extract_dHdl(outfile):
+    """Return gradients `dH/dl` from Amebr TI outputfile
+    Parameters
+    ----------
+    outfile : str
+        Path to Amber .out file to extract data from.
+
+    Returns
+    -------
+    dH/dl : Series
+        dH/dl as a function of time for this lambda window.
+    """
     file_datum = file_validation(outfile)
-    if file_validation(outfile):
-        finished = False
-        comps = []
-        with SectionParser(outfile) as secp:
-            line = secp.skip_lines(5)
-            nensec = 0
-            nenav = 0
-            old_nstep = -1
-            old_comp_nstep = -1
-            high_E_cnt = 0
-
-            in_comps = False
-            for line in secp:
-                if 'DV/DL, AVERAGES OVER' in line:
-                    in_comps = True
-                if line.startswith(' NSTEP'):
-                    if in_comps:
-                        #CHECK the result
-                        result = secp.extract_section('^ NSTEP', '^ ---',
-                                                     ['NSTEP'] + DVDL_COMPS,
-                                                     extra=line)
-                        if result[0] != old_comp_nstep and not any_none(result):
-                            comps.append([float(E) for E in result[1:]])
-                            nenav += 1  
-                            old_comp_nstep = result[0]
-                        in_comps = False
-                    else:
-                        nstep, dvdl = secp.extract_section('^ NSTEP', '^ ---',
-                                                           ['NSTEP', 'DV/DL'],
-                                                           extra=line)
-                        if nstep != old_nstep and dvdl is not None \
-                                and nstep is not None:
-                            file_datum.gradients.append(dvdl)
-                            nensec += 1
-                            old_nstep = nstep
-                if line == '   5.  TIMINGS\n':
-                    finished = True
-                    break
-        if not finished:
-            print('  WARNING: prematurely terminated run')
-        if not nensec:
-            print('  WARNING: File %s does not contain any DV/DL data\n' %
-                  outfile)
-        print('%i data points, %i DV/DL averages' % (nensec, nenav))
-        #at this step we get info stored in the FEData object for a given amber out file
-        file_datum.component_gradients.extend(comps)
-        #convert file_datum to the pandas format to make it identical to alchemlyb output format
-        df = convert_to_pandas(file_datum)        
-    else:
-        df = None
+    if not file_validation(outfile):
+        return None
+    finished = False
+    comps = []
+    with SectionParser(outfile) as secp:
+        line = secp.skip_lines(5)
+        nensec = 0
+        nenav = 0
+        old_nstep = -1
+        old_comp_nstep = -1
+        high_E_cnt = 0
+        in_comps = False
+        for line in secp:
+            if 'DV/DL, AVERAGES OVER' in line:
+                in_comps = True
+            if line.startswith(' NSTEP'):
+                if in_comps:
+                    #CHECK the result
+                    result = secp.extract_section('^ NSTEP', '^ ---',
+                                                 ['NSTEP'] + DVDL_COMPS,
+                                                 extra=line)
+                    if result[0] != old_comp_nstep and not any_none(result):
+                        comps.append([float(E) for E in result[1:]])
+                        nenav += 1  
+                        old_comp_nstep = result[0]
+                    in_comps = False
+                else:
+                    nstep, dvdl = secp.extract_section('^ NSTEP', '^ ---',
+                                                       ['NSTEP', 'DV/DL'],
+                                                       extra=line)
+                    if nstep != old_nstep and dvdl is not None \
+                            and nstep is not None:
+                        file_datum.gradients.append(dvdl)
+                        nensec += 1
+                        old_nstep = nstep
+            if line == '   5.  TIMINGS\n':
+                finished = True
+                break
+    if not finished:
+        logging.warning('  WARNING: prematurely terminated run')
+    if not nensec:
+        logging.warning('  WARNING: File %s does not contain any DV/DL data\n' %
+              outfile)
+    logging.info('%i data points, %i DV/DL averages' % (nensec, nenav))
+    #at this step we get info stored in the FEData object for a given amber out file
+    file_datum.component_gradients.extend(comps)
+    #convert file_datum to the pandas format to make it identical to alchemlyb output format
+    df = convert_to_pandas(file_datum)        
     return df
 
 #currently just check the code with a simple amber ti output file
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -8,6 +8,7 @@ import re
 import pandas as pd
 import numpy as np
 import logging 
+from .util import anyopen
 
 logger = logging.getLogger("alchemlyb.parsers.Amber")
 
@@ -61,16 +62,8 @@ class SectionParser(object):
     def __init__(self, filename):
         """Opens a file according to its file type."""
         self.filename = filename
-        with open(filename, 'r') as f:
-            magic = f.read(3)   # NOTE: works because all 3-byte headers
         try:
-            method = _MAGIC_CMPR[magic]
-        except KeyError:
-            open_it = open
-        else:
-            open_it = getattr(__import__(method[0]), method[1])
-        try:
-            self.fileh = open_it(self.filename, 'r')
+            self.fileh = anyopen(self.filename, 'r')
             self.filesize = os.stat(self.filename).st_size
         except Exception as ex:
             logging.exception("ERROR: cannot open file %s" % filename)
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -1,0 +1,265 @@
+"""Parsers for extracting alchemical data from amber output files.
+Most of the file parsing part are inheriting from alchemical-analysis  
+Change the final format to pandas to be consistent with the alchemlyb format
+"""
+
+import pandas as pd
+import re
+import numpy as np
+import os
+
+def convert_to_pandas(file_datum, ):
+    data_dic = {}
+    data_dic["dHdl"] = []
+    data_dic["lambdas"] = []
+    data_dic["time"] = []
+    for frame_index, frame_dhdl in enumerate(file_datum.gradients):
+        data_dic["dHdl"].append(frame_dhdl)
+        data_dic["lambdas"].append(file_datum.clambda)
+        #here we need to convert dt to ps unit from ns 
+        frame_time = file_datum.t0 + (frame_index + 1) * file_datum.dt*1000
+        data_dic["time"].append(frame_time)
+    df = pd.DataFrame(data_dic["dHdl"], columns=["dHdl"], index =pd.Float64Index(data_dic["time"], name='time'))
+    df["lambdas"] = data_dic["lambdas"][0]
+    df = df.reset_index().set_index(['time'] + ['lambdas'])
+    return df
+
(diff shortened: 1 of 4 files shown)


davidlmobley:
@shuail - this will need to come with tests. You'll probably want to put the data for the tests in `alchemtest` and then build in tests which rely on it here.

Note also that there is a "code coverage" analysis being run above which I presume is checking to see what portion of the code is being tested by the tests; since yours doesn't include tests, well, code coverage reduces dramatically and fails that test.

orbeckst:
As just discussed: 
1. add tests to alchemtest (open a PR there)
2. write tests (initially modelled on the Gromacs ones) that use the alchemtest data set

Ping me if _anything_ is unclear.

shuail:
@orbeckst just added the amber dataset in alchemtest and create a PR there. Added amber test here to use the files in alchemtest. Tested on my local computer to use the updated alchemtest, it runs fine. Finger crossed for travis once the alchemtest library get updated.

orbeckst:
I'll deal with PR  alchemistry/alchemtest#6 first to sort out the data side of things.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
+"""Parsers for extracting alchemical data from amber output files.
+Most of the file parsing part are inheriting from alchemical-analysis  
+Change the final format to pandas to be consistent with the alchemlyb format
+"""
+
+import pandas as pd
+import re
+import numpy as np
+import os
+
```
orbeckst: Use [logging](https://docs.python.org/3/library/logging.html) instead of `print`:
```
import logging

logger = logging.getLogger("alchemlyb.parsers.Amber")
```

See more comments below.

Note that in order to see the messages, a root logger has to be created elsewhere:
```python
import logging

def create_alchemlyb_logger(logfile="alchemlyb.log", name="alchemlyb"):
    """Create a logger that outputs to screen and to `logfile`"""

    logger = logging.getLogger(name)
    logger.setLevel(logging.DEBUG)

    # handler that writes to logfile
    logfile_handler = logging.FileHandler(logfile)
    logfile_formatter = logging.Formatter('%(asctime)s %(name)-12s %(levelname)-8s %(message)s')
    logfile_handler.setFormatter(logfile_formatter)
    logger.addHandler(logfile_handler)

    # define a Handler which writes INFO messages or higher to the sys.stderr
    console_handler = logging.StreamHandler()
    console_handler.setLevel(logging.INFO)
    # set a format which is simpler for console use
    formatter = logging.Formatter('%(name)-12s: %(levelname)-8s %(message)s')
    console_handler.setFormatter(formatter)
    logger.addHandler(console_handler)

    return logger

rootlogger = create_alchemlyb_logger()
```
We might add this convenience function elsewhere in the library. I opened issue #34. 

Note that you only have to create the top level logger (here called `rootlogger`) once. Loggers are known globally in the interpreter, so you can "attach" to the root logger from anywhere else just by naming the new logger "alchemlyb.OTHER.STUFF".

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+            open_it = open
+        else:
+            open_it = getattr(__import__(method[0]), method[1])
+        try:
+            self.fileh = open_it(self.filename, 'rb')
+            self.filesize = os.stat(self.filename).st_size
+        except IOError:
+            raise SystemExit('ERROR: cannot open file %s' % filename)
+        self.lineno = 0
+    def skip_lines(self, nlines):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        """Skip a given number of files."""
+        lineno = 0
+        for line in self:
+            lineno += 1
+            if lineno > nlines:
+                return line
+        return None
+    def skip_after(self, pattern):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        """Skip until after a line that matches a regex pattern."""
+        for line in self:
+            match = re.search(pattern, line)
+            if match:
+                break
+        return self.fileh.tell() != self.filesize
+    def extract_section(self, start, end, fields, limit=None, extra='',
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+                    result.append(float('Inf') )
+                # NOTE: check if this is a sufficient test for int
+                elif '.' not in value and re.search(r'\d+', value):
+                    result.append(int(value))
+                else:
+                    result.append(float(value))
+            else:                       # section may be incomplete
+                result.append(None)
+        return result
+    def __iter__(self):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        return self
+    def next(self):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        """Read next line of the filehandle and check for EOF."""
+        self.lineno += 1
+        curr_pos = self.fileh.tell()
+        if curr_pos == self.filesize:
+            raise StopIteration
+        # NOTE: can't mix next() with seek()
+        return self.fileh.readline()
+    def close(self):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        """Close the filehandle."""
+        self.fileh.close()
+    def __enter__(self):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        return self
+    def __exit__(self, typ, value, traceback):
```
orbeckst: [PEP8](https://www.python.org/dev/peps/pep-0008/) formatting: leave a space between methods.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        self.close()
+
+class FEData(object):
+    """A simple struct container to collect data from individual files."""
+
+    #__slots__ = ['clambda', 't0', 'dt', 'T', 'gradients',
```
orbeckst: remove commented out code if not necessary

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+        self.component_gradients = []
+        #self.mbar_energies = []
+
+
+def file_validation(outfile, ):
+    invalid = False
+    with SectionParser(outfile) as secp:
+        line = secp.skip_lines(5) 
+        if not line:
+            print('  WARNING: file does not contain any useful data, '
```
orbeckst: turn `print` into 
```python
logger.warn("file %r does not contain any useful data, skipping file", outfile)
```
Add the filename to the error message; I assume that `outfile` was a filename.

orbeckst: If you think that users should always see the message, use `warnings.warn(msg)` in addition to the logger.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+                      'ignoring file')
+            invalid = True
+        if not secp.skip_after('^   2.  CONTROL  DATA  FOR  THE  RUN'):
+            print('  WARNING: no CONTROL DATA found, ignoring file')
```
orbeckst: use logger

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+            invalid = True
+        ntpr, = secp.extract_section('^Nature and format of output:', '^$',
+                                     ['ntpr'])
+        nstlim, dt = secp.extract_section('Molecular dynamics:', '^$',
+                                          ['nstlim', 'dt'])
+        T, = secp.extract_section('temperature regulation:', '^$',
+                                 ['temp0'])
+        if not T:
+            raise SystemExit('ERROR: Non-constant temperature MD not '
```
orbeckst: Do not `SystemExit` in a library.

This is a library and calling code might want to handle it. Raise `NotImplementedError` instead.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+                             'currently supported')
+            invalid = True
+        clambda, = secp.extract_section('^Free energy options:', '^$',
+                                        ['clambda'], '^---')
+        if clambda is None:
+            print('  WARNING: no free energy section found, ignoring file')
```
orbeckst: use logging

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+            invalid = True
+
+        if not secp.skip_after('^   3.  ATOMIC '):
+            print('  WARNING: no ATOMIC section found, ignoring file\n')
```
orbeckst: use logging

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+            invalid = True
+
+        t0, = secp.extract_section('^ begin time', '^$', ['coords'])
+        if not secp.skip_after('^   4.  RESULTS'):
+            print('  WARNING: no RESULTS section found, ignoring file\n')
```
orbeckst: use logging

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+            invalid = True
+    if invalid:
+        return False
+    else:
+        file_datum = FEData()
+        file_datum.clambda = clambda
+        file_datum.t0 = t0
+        file_datum.dt = dt
+        file_datum.T = T
+        return file_datum
```
orbeckst: I would simplify to
```python
if invalid:
        return False

file_datum = FEData()
file_datum.clambda = clambda
file_datum.t0 = t0
file_datum.dt = dt
file_datum.T = T
return file_datum
```

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,265 @@
...
+
+def extract_dHdl(outfile, ):
+    file_datum = file_validation(outfile)
+    if file_validation(outfile):
```
orbeckst: rewrite the `if` to terminate early, the following would be clearer:
```python
if not file_validation(outfile):
   return None

finished = False
comps = []
...
return df
```

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 from pymbar.timeseries import statisticalInefficiency
 from pymbar.timeseries import detectEquilibration
+from pymbar.timeseries import subsampleCorrelatedData
```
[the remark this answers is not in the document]
orbeckst: I am not reviewing these changes here because they will be in a separate PR.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -0,0 +1,21 @@
+"""Amber parser tests.
+
+"""
+
+from alchemlyb.parsing.amber import extract_dHdl
+from alchemtest.amber import load_simplesolvated
+
+
+def test_dHdl():
```
orbeckst: good start

src/alchemlyb/parsing/amber.py
shuail: Yes, will switch to logging instead of printing. 

Just want to make sure that I understand the logic correctly, so what I need to do is to add 

```
import logging

logger = logging.getLogger("alchemlyb.parsers.Amber")
```

at the beginning of the code and start to logging like

logging.info ("Write some logging info here") is that correct?

orbeckst: Yes, exactly.

(Just be aware that until a logger named "alchemlyb" has been created, no logging will be visible, but that's expected and desired behavior. With #34 I will add code to the library to start logging automatically.)

shuail:
@orbeckst Thanks for all your comments, it's very helpful. I made three major changes 1, switch to logging in the amber parser and clean up the code (add docs, empty lines etc). 2, add test_ti_estimators_amber.py to test the ti_estimator for amber. 3, roll back the subsampling commit and will create a separated PR later.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,282 @@
...
+        logging.warning('  WARNING: File %s does not contain any DV/DL data\n' %
+              outfile)
+    logging.info('%i data points, %i DV/DL averages' % (nensec, nenav))
+    #at this step we get info stored in the FEData object for a given amber out file
+    file_datum.component_gradients.extend(comps)
+    #convert file_datum to the pandas format to make it identical to alchemlyb output format
+    df = convert_to_pandas(file_datum)        
+    return df
+
+#currently just check the code with a simple amber ti output file
```
orbeckst: remove the whole main section  (also, the `print` use python 2.7 syntax but we run under python 2 and python 3).

shuail: Got it, the main section got removed now.

orbeckst:
This is starting to look good.

We would now like to get test coverage up. If you look at https://codecov.io/gh/alchemistry/alchemlyb/src/master/src/alchemlyb/parsing/amber.py (follow the codecov/project report and look for the overall coverage and navigate in the files) there are many lines in the function [parsing.amber.file_validation()](https://codecov.io/gh/alchemistry/alchemlyb/src/master/src/alchemlyb/parsing/amber.py#L171) that are not exercised by the tests. We would want to have tests that actually show that you catch these errors as expected. 

The best thing to do would be to include small files with errors and then just test the `file_validation()` function itself on these files. It is difficult to check that a log message was printed but you can test that `file_validation()` returns `False` for these cases. (As a sidenote, if you need to [test exceptions and warnings](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) then pytest also makes this straight forward).

(This is where finalizing PRs becomes painful and boring but it is vital for the long-term survivability of the code that we check things *now* and are then assured that in the future everything will still work *or* we learn when it doesn't.)

Other comments on the code.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,277 @@
...
+        yield it.next()
+
+class SectionParser(object):
+    """
+    A simple parser to extract data values from sections.
+    """
+    def __init__(self, filename):
+        """Opens a file according to its file type."""
+        self.filename = filename
+        with open(filename, 'r') as f:
```
orbeckst: This approach with `_MAGIC_CMPR` can be replace with `util.anyopen()`, see, for example https://github.com/alchemistry/alchemlyb/blob/master/src/alchemlyb/parsing/gmx.py#L143 . That's cleaner and more expandable. (I hadn't picked up on it in my earlier code reviews.)

(You can do it yourself or I am also happy to finish the PR as is and do this later and separately.)

shuail:
@orbeckst Got it, will add some testing for the file validation function. For the tested output files, I planned to created an invalidfiles folder under alchemtest/amber and use the alchemtest frame work in the alchemlyb. Does this sound good?

src/alchemlyb/parsing/amber.py
shuail: Got it, will try to switch to anyopen

orbeckst:
Yes, good idea. Just check that the files are properly included when you run `pip install`, i.e., proper inclusions in `setup.py` (and perhaps `MANIFEST.in`... if we use it).

shuail:
@orbeckst Just finished adding the test code here and test files (alchemtest) to test the file validation function in amber parser and switched the file opener to anyopen. Hope the test will pass this time :)

orbeckst:
Looking good, just waiting for PR alchemistry/alchemtest#11.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
...
+        return False
+    file_datum = FEData()
+    file_datum.clambda = clambda
+    file_datum.t0 = t0
+    file_datum.dt = dt
+    file_datum.T = T
+    return file_datum
+
+def extract_dHdl(outfile):
+    """Return gradients `dH/dl` from Amebr TI outputfile
```
orbeckst: typo: Amber

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
...
+class SectionParser(object):
+    """
+    A simple parser to extract data values from sections.
+    """
+    def __init__(self, filename):
+        """Opens a file according to its file type."""
+        self.filename = filename
+        try:
+            self.fileh = anyopen(self.filename, 'r')
+            self.filesize = os.stat(self.filename).st_size
```
orbeckst: This cannot be used for compressed files (see other comments).

I think this line can be deleted because it is useless.

orbeckst: Well, thinking about this again: try it out, maybe it works and I am wrong.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
...
+                return line
+        return None
+
+    def skip_after(self, pattern):
+        """Skip until after a line that matches a regex pattern."""
+        for line in self:
+            match = re.search(pattern, line)
+            if match:
+                break
+        return self.fileh.tell() != self.filesize
```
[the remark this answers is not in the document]
orbeckst: _Well, thinking about this again: try it out, maybe it works and I am wrong._

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
+"""Parsers for extracting alchemical data from amber output files.
+Most of the file parsing part are inheriting from alchemical-analysis  
+Change the final format to pandas to be consistent with the alchemlyb format
+"""
+
+import os
+import re
+import pandas as pd
+import numpy as np
+import logging
```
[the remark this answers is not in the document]
shuail: Got it, push a fixed

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
...
+        data_dic["time"].append(frame_time)
+    df = pd.DataFrame(data_dic["dHdl"], columns=["dHdl"], index =pd.Float64Index(data_dic["time"], name='time'))
+    df["lambdas"] = data_dic["lambdas"][0]
+    df = df.reset_index().set_index(['time'] + ['lambdas'])
+    return df
+
+DVDL_COMPS = ['BOND', 'ANGLE', 'DIHED', '1-4 NB', '1-4 EEL', 'VDWAALS',
+              'EELEC', 'RESTRAINT']
+_FP_RE = r'[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?'
+_MAGIC_CMPR = {
```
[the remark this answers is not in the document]
shuail: removed

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
...
+    return False
+
+def _pre_gen(it, first):
+    """A generator that returns first first if it exists."""
+
+    if first:
+        yield first
+
+    while it:
+        yield it.next()
```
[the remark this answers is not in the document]
shuail: Yes, change it to next(it)

src/alchemlyb/parsing/amber.py
shuail: Yes, I think you are right, the filesize is not working here for zipped files, remove the filesize related detection in the code.

src/alchemlyb/parsing/amber.py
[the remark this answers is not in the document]
shuail: Yes, I checked that the `self.filesize < self.fileh.tell()` for the zipped file and it will not be equal at the end of the file so I changed the `skip_after()` function to not use this detection but explicitly return `True` if the pattern was found along iterating the whole file

orbeckst:
I merged alchemistry/alchemtest#11 so you should now have the appropriate test files available for Travis CI. I will manually restart the tests. You don't need to do anything until the test results come back.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,270 @@
...
+        return result
+
+    def __iter__(self):
+        return self
+
+    def next(self):
+        """Read next line of the filehandle and check for EOF."""
+        self.lineno += 1
+        curr_pos = self.fileh.tell()
+        if curr_pos == self.filesize:
```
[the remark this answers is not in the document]
shuail: Yes, that's a better solution. Sorry I didn't think that part much as migrating the code from alchemical-analysis. It looks like that this test successfully catch the untested part of the original code!

src/alchemlyb/parsing/amber.py
shuail: Fixed

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,264 @@
...
+Change the final format to pandas to be consistent with the alchemlyb format
+"""
+
+import os
+import re
+import logging 
+import pandas as pd
+import numpy as np
+
+from util import anyopen
```
orbeckst: Needs to be 
```python
from .util import anyopen
```

shuail:
@orbeckst thanks for your comments! Yes, after I switched to the zipped files, the end of file(EOF) detection in the `skip_after` function will not work, so I changed that part and also the next() function to make it suitable for the zipped file. The `anyopen` actually works fine here.

src/alchemlyb/parsing/amber.py
orbeckst: https://travis-ci.org/alchemistry/alchemlyb/jobs/301541570

orbeckst:
One small change to get it to pass under python 3.

src/alchemlyb/parsing/amber.py
shuail: Oops, yes, fixed this

orbeckst:
Once the tests pass, we have the coverage. Assuming that this will look good, we need to make sure that it can be merged and I want to clean up the PR, i.e., merging related commits and removing the subsampling code that was added and deleted. I can do this locally with `git rebase`. 

A consequence for you will be that you also have to rebase your fork because you have been working from master instead of a branch. I can tell you what I would do if you want to know.

orbeckst:
This looking very good. The only way that you can sensibly improve the coverage is by adding a test for the function
```python
parsing.amber.any_none(sequence)
```
Basically, make sure that the [line 42](https://codecov.io/gh/alchemistry/alchemlyb/compare/8915ece9daa2ad17094b2e61565bcfe128a374c7...49cc45cf3f310e1a6eff1d4c15f6fa2c17adbd09/diff#D1-42) is hit. Any of the other remaining checks are fairly standard defensive programming and I am not even sure how to test them properly.

src/alchemlyb/parsing/amber.py
```diff
@@ -0,0 +1,264 @@
...
+DVDL_COMPS = ['BOND', 'ANGLE', 'DIHED', '1-4 NB', '1-4 EEL', 'VDWAALS',
+              'EELEC', 'RESTRAINT']
+_FP_RE = r'[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?'
+
+def any_none(sequence):
+    """Check if any element of a sequence is None."""
+
+    for element in sequence:
+        if element is None:
+            return True
```
orbeckst: This looking very good. The only way that you can sensibly improve the coverage is by adding a test for the function
```python
parsing.amber.any_none(sequence)
```
Basically, make sure that the [line 42](https://codecov.io/gh/alchemistry/alchemlyb/compare/8915ece9daa2ad17094b2e61565bcfe128a374c7...49cc45cf3f310e1a6eff1d4c15f6fa2c17adbd09/diff#D1-42) is hit. Any of the other remaining checks are fairly standard defensive programming and I am not even sure how to test them properly.

shuail: Got a test function added for this line

orbeckst:
@shuail looking good, I'll do the rebase... hopefully done before I have to run to class...

orbeckst:
I created PR #36 to finalize.

PR #35, opened 2017-11-11
use versioneer for automated versioning

### LINKED ISSUE #30: add __version__ or other version information
I have not found a way to ask `alchemlyb` which version it is. Something like
```python
alchemlyb.__version__
```
would be useful.

Maybe use [versioneer](https://github.com/warner/python-versioneer) to be fancy and hip.

orbeckst:
Closes #30.

(Uses https://github.com/warner/python-versioneer with PEP440 style version information)


dotsdl:
@orbeckst this is pretty slick. I haven't used `versioneer` for any projects before, but this looks like a solution for simplifying the (always tedious) release process. It even works when the current commit isn't tagged, grabbing what looks like the first tag it finds in the branch and appending a number indicating the commit.

I tried it out myself, and it appears to work as advertised. I did:
1. `git tag 0.2.0`
2. `python setup.py sdist`
3. `mkvirtualenv alchemlyb-vers`
4. `workon alchemlyb-vers`
5. `pip install numpy six` # for pymbar installation
6. `pip install dist/alchemlyb-0.2.0.tar.gz`
7. `cd ~ ; python -c "import alchemlyb; print(alchemlyb.__version__)"

And we get '0.2.0' out of this as expected. Very nice!

davidlmobley:
@nathanmlim possibly of interest

PR #38, opened 2017-11-14
change the subsampling method 

shuail:
change the subsampling code to use the pymbar module to get the subsampled indexes, the original code use the slicing function in pandas which requires the slicing step to be integer. While in theory I think we don't want to force the subsampling steps to be integer, will open an issue to discuss this.

### DIFF
--- a/src/alchemlyb/preprocessing/subsampling.py
+++ b/src/alchemlyb/preprocessing/subsampling.py
@@ -4,6 +4,7 @@
 import numpy as np
 from pymbar.timeseries import statisticalInefficiency
 from pymbar.timeseries import detectEquilibration
+from pymbar.timeseries import subsampleCorrelatedData
 
 
 def _check_multiple_times(df):
@@ -97,14 +98,16 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None)
 
         # calculate statistical inefficiency of series
         statinef  = statisticalInefficiency(series)
+        
+        #use the subsampleCorrelatedData function to get the subsample index
+        indices = subsampleCorrelatedData(series, g=statinef)
+        picked_time_index = []
+        #pick the time index for the pandas dataframe based on the python index from subsample
+        for s_index, s_index_pair in enumerate(series.index):
+            if s_index in indices:
+                picked_time_index.append(s_index_pair[0])
 
-        # we round up
-        statinef = int(np.rint(statinef))
-
-        # subsample according to statistical inefficiency
-        series = series.iloc[::statinef]
-
-        df = df.loc[series.index]
+        df = df.loc[picked_time_index]
     else:
         df = slicing(df, lower=lower, upper=upper, step=step)
     
--- a/src/alchemlyb/preprocessing/subsampling.py
+++ b/src/alchemlyb/preprocessing/subsampling.py
@@ -2,9 +2,7 @@
 
 """
 import numpy as np
-from pymbar.timeseries import statisticalInefficiency
-from pymbar.timeseries import detectEquilibration
-from pymbar.timeseries import subsampleCorrelatedData
+from pymbar.timeseries import statisticalInefficiency, detectEquilibration, subsampleCorrelatedData
 
 
 def _check_multiple_times(df):
@@ -101,13 +99,7 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None)
         
         #use the subsampleCorrelatedData function to get the subsample index
         indices = subsampleCorrelatedData(series, g=statinef)
-        picked_time_index = []
-        #pick the time index for the pandas dataframe based on the python index from subsample
-        for s_index, s_index_pair in enumerate(series.index):
-            if s_index in indices:
-                picked_time_index.append(s_index_pair[0])
-
-        df = df.loc[picked_time_index]
+        df = df.iloc[indices]
     else:
         df = slicing(df, lower=lower, upper=upper, step=step)
     


src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 from pymbar.timeseries import statisticalInefficiency
 from pymbar.timeseries import detectEquilibration
+from pymbar.timeseries import subsampleCorrelatedData
```
orbeckst: import together (combine all three lines, no idea why they were separate)
```python
from pymbar.timeseries import statisticalInefficiency, detectEquilibration, subsampleCorrelatedData
```

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -97,14 +98,16 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None)
...
-        # we round up
-        statinef = int(np.rint(statinef))
-
-        # subsample according to statistical inefficiency
-        series = series.iloc[::statinef]
-
-        df = df.loc[series.index]
+        
+        #use the subsampleCorrelatedData function to get the subsample index
+        indices = subsampleCorrelatedData(series, g=statinef)
```
orbeckst: Can you briefly explain what `subsampleCorrelatedData` does (and maybe link to the code in their repo)?

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -97,14 +98,16 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None)
...
+        picked_time_index = []
+        #pick the time index for the pandas dataframe based on the python index from subsample
+        for s_index, s_index_pair in enumerate(series.index):
```
orbeckst: This looks clunky and potentially slow (although that might not matter much). Isn't there a direct way to do this without the `for` loop?

(I don't really know what the data look like to say more here.)

src/alchemlyb/preprocessing/subsampling.py
shuail: @orbeckst yes, the link to the code is https://github.com/choderalab/pymbar/blob/ac648e63787592a251992a764876713a64f17455/pymbar/timeseries.py#L632 it basically subsample the time series (`series`) based on the given `g` (statistical inefficiency)

src/alchemlyb/preprocessing/subsampling.py
shuail: Yea, let me check if there is a better way to rewrite this

src/alchemlyb/preprocessing/subsampling.py
dotsdl: I don't quite see the point of this change. From what I can tell the `subsampleCorrelatedData` function [deals in integer-spaced samples, too](https://github.com/choderalab/pymbar/blob/ac648e63787592a251992a764876713a64f17455/pymbar/timeseries.py#L715). @shuail can you explain how this approach really differs from the one we already have here?

src/alchemlyb/preprocessing/subsampling.py
dotsdl: Yes, this will be very slow. You generally want to avoid looping through the rows in a `DataFrame` or `Series`. You can directly extract all values for an integer-based index with ``series.iloc[indices]`` instead.

src/alchemlyb/preprocessing/subsampling.py
shuail: @dotsdl yes, the code you refer to is an optional option `conservative`, and the default for that option is `False`. I didn't notice that option before while looking at the details, I think the implementation is also different from what alchemlyb currently had. 

The topic here is basically all related to the precision and distinguishability of the subsampling. For a given `g` like 1.5, current alchemlyb code will round up using `int(np.rint(1.5))` = 1.0 and, the `conservative` option will round to 2 using `math.ceil(1.5)` and the `non conservative` option will use 1.5. For a given time series with 1,000 data point, current alchemlyb; `conservative`; `non conservative`  will get 1,000; 500; 666 data point respectively. Since the energy will depend on the subsampling, I think we should be careful about how we round the `g` and I suggested to use the `non conservative` option here to make all dataset as much distinguishable as possible.

src/alchemlyb/preprocessing/subsampling.py
shuail: @dotsdl yes, the `iloc[indices]` works. Thx!

src/alchemlyb/preprocessing/subsampling.py
shuail: @orbeckst yes, changed to one line

src/alchemlyb/preprocessing/subsampling.py
orbeckst: Add explicit keywords for `subsampleCorrelatedData(..., fast=False, conservative=False, verbose=False)` (especially `conservative=False` to make sure that the behavior in alchemlyb won't change, even if pymbar changes its defaults.)

orbeckst: I agree with https://github.com/alchemistry/alchemlyb/pull/38#discussion_r150987109 – until we come up with anything better inside alchemlyb I would also go with the pymbar version.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -97,14 +96,10 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None)
...
-
-        # we round up
-        statinef = int(np.rint(statinef))
-
-        # subsample according to statistical inefficiency
-        series = series.iloc[::statinef]
-
-        df = df.loc[series.index]
+        
+        #use the subsampleCorrelatedData function to get the subsample index
```
orbeckst: update the docs: make clear that intervals may be not uniform any more.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -97,14 +96,10 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None)
...
+        indices = subsampleCorrelatedData(series, g=statinef)
```
orbeckst: Add explicit keywords for `subsampleCorrelatedData(..., fast=False, conservative=False, verbose=False)` (especially `conservative=False` to make sure that the behavior in alchemlyb won't change, even if pymbar changes its defaults.)

orbeckst: Have we got tests that test fractional `g`?

orbeckst:
Btw, @shuail this PR appears to come from your master branch. I recommend that [PRs get their own branches](https://github.com/alchemistry/alchemlyb/wiki/How-to-do-PRs#prs-get-their-own-branches).

src/alchemlyb/preprocessing/subsampling.py
shuail: @orbeckst yes, sounds good to me. I will add the default options explicitly.

src/alchemlyb/preprocessing/subsampling.py
shuail: yes, will add docs

src/alchemlyb/preprocessing/subsampling.py
shuail: There is a test_preprocessing.py code, and inside there is a test_subsampling function https://github.com/alchemistry/alchemlyb/blob/0587ebd384b47eeb16e0d4c7acb1029c03b7ddda/src/alchemlyb/tests/test_preprocessing.py#L76 testing that the subsampled data is not bigger than the original data, while I think this one didn't explicitly test the `g` value or the actual size of the subsample data. Also, should we define a slicer function inside this class, maybe a bug here?

shuail:
@orbeckst thanks for the comments, yes, will clean up the code (add more docs/explicit the options) and create another PR from the not master branch.

davidlmobley:
So just to chime in here -- we're talking about a fairly significant change on the basis of how to round `g`, is that correct? Specifically:

> The topic here is basically all related to the precision and distinguishability of the subsampling. For a given g like 1.5, current alchemlyb code will round up using int(np.rint(1.5)) = 1.0 and, the conservative option will round to 2 using math.ceil(1.5) and the non conservative option will use 1.5. For a given time series with 1,000 data point, current alchemlyb; conservative; non conservative will get 1,000; 500; 666 data point respectively. Since the energy will depend on the subsampling, I think we should be careful about how we round the g and I suggested to use the non conservative option here to make all dataset as much distinguishable as possible.

Let's think for a minute about the practical difference between using different values of `g`. When will this be important? I think in the limit of "well behaved" data, where we have plenty of statistically distinguishable samples, this is not going to really matter at all, so the change is unimportant and is unlikely to buy us any real value; presumably it will make uncertainty estimates a little lower, but that's not especially valuable. When it will be important is closer to the limit of "problematic" data, where we may not have enough sampling at all, or we just barely have enough sampling to get a reasonable estimate. This is the case we really want to be thinking about -- maybe $N_{eff}$ is 20 or 50 so we only have a rather small number of samples at some lambda value.  

In that case -- as we approach the limit of problematic data -- what's our biggest goal? Is it being able to squeeze every last bit of POSSIBLE value out of the data we have, even if that results in a false appearance of precision when in fact our results are garbage? Or is it being able to correctly recognize that in fact we don't have enough data? I'd argue that the latter goal is more important: It's worse to get an "apparently reasonable" answer that is wrong than to get a "clearly noisy/useless" answer.

Thus, in my view, we should err in the direction of conservatism (at least by default) and round `g` up  if we need to round. I could be OK with not rounding (though see below) but certainly not rounding down.

In my previous experience though, statistical uncertainties tend to always be underestimated anyway, so rounding up seems like a relatively safe thing to do if we need to round.

So in my mind the decision is between:
- round up (and keep current code)
- go with non-conservative option and just use numeric value of g, potentially giving non-uniform spaced frames

One of our overall goals is to allow easy coupling with things like trajectory analysis, which may be an argument in favor of the simpler approach (round g upwards to the nearest integer) since having non-uniform frame spacing is going to complicate the analysis a lot.

shuail:
@davidlmobley In our practice, supposing `g` will be linearly correlated to the `subsample_diff` which is = `energy(subsampled) - energy(not subsampled)` and we want to use the `subsample_diff` as an indicator for other further analysis ie, filtering the dataset, machine learning etc. It will be nice to have that indicator behave linearly so that `g` is not rounded. While I can see that round up will not effect the detecting of large `subsample_diff` which saying that `g` = 30 vs 31 will not have much difference, so I suggest that we adopt the `subsampleCorrelatedData` option here (`conservative=False` as default) and add a `conservative` flag to pass to `subsampleCorrelatedData`. Another thing that I am thinking to avoid using `round up` as default is that for cases like `g` = 1.01, the rounded up `g` will be 2 and it turns out that we just pick half of the data even they are highly uncorrelated. 

For coupling the energy vs trajectory analysis, I think the pacing of writing out energy vs trajectory is arbitrary and will depend on the users input, even the uniform frame spacing of subsampling would not benefit much in this case. So I would like to use the unsubsampled data for the energy vs trajectory analysis.

orbeckst:
Good points. 

The current code rounds to nearest integer so that can create the appearance of more independent samples than there really are. 

Perhaps use the conservative approach as default and add an option for users to choose the non-uniform sampling (and making sure that the frame indices for slicing the original trajectory remain available downstream).

orbeckst:
To summarize the discussion so far:
- We rather want to be conservative so the *current implementation* using `statinef = int(np.rint(statinef))` **should be changed** because [numpy.rint](https://docs.scipy.org/doc/numpy/reference/generated/numpy.rint.html) can round down (e.g., `g=1.4` will give `g=1`). If nothing else happens it should become `statinef = int(np.ceil(statinef))`.
- We could completely rely on [`pymbar.timeseries.subsampleCorrelatedData()`](https://github.com/choderalab/pymbar/blob/ac648e63787592a251992a764876713a64f17455/pymbar/timeseries.py#L632) and use the `conservative=True` keyword to get the `ceil(g)` rounding.
- If we use `pymbar.timeseries.subsampleCorrelatedData()`then we can also use `conservative=False` to obtain indices with uneven sampling that best approximate a fractional `g`. I don't think that unevenly sampled frames will make any difference for downstream processing in alchemlyb. It might become more difficult to use the processed energies in the context of a trajectory but this might be less of a concern because
  - energy and trajectory output are often not written out at the same frequency  and slicing one by the other is tricky business anyway that has to be solved elsewhere (e.g. by comparing time stamps of frames – our [standard forms of raw data](http://alchemlyb.readthedocs.io/en/latest/parsing.html#standard-forms-of-raw-data) contain time stamps, so they can be used)
  -  a more typical approach might be (??) to featurize a trajectory, from the featurization determine how to slice the *raw* energy time series, and then subsample the sliced time series according to the `g` of the raw time series – in this case, we never need to care about even or uneven sampling.

I suggest that we 
- add a keyword to allow the user to switch between `conservative=True` and `False` (I don't like the kw name very much but we should keep it to be consistent with pymbar)
- use `pymbar.timeseries.subsampleCorrelatedData()` for fractional `g` or `ceil(g)`

The main question that would remain is **what should the default be**?

I am tending towards `conservative=True` to protect unsuspecting users from overly optimistic estimates while still allowing knowledgable users to get what they want.

davidlmobley:
I think I'm on board with all of this and also lean towards having the default be `conservative=True`, for the same reasons just noted.

shuail:
I think I'm fine with @orbeckst suggested procedure. For the default option, I am also fine with the rounding approach since that's close to the original code and philosophy. I could change the option for my own use anyway.

orbeckst:
Good, so we have a consensus. 

Can you please modify your PR accordingly?

We will also need at least two tests that check that the two approaches work. 

--
Oliver Beckstein
email: orbeckst@gmail.com

shuail:
Yes, I will work on the code and the tests. Once I finished that, will create another PR from a separate branch other than `master`

orbeckst:
I continued the work on PR #41 – please push updates there. (#41 was rebased against master!)

PR #43, opened 2017-12-01
issue-14-expanded-ensemble-parser

trje3733:
Modified gmx parser to handle expanded ensemble simulation data. Original parser relies on the state being disclosed in the subtitle line of the dhdl.xvg, which is not the case in expanded ensemble datasets. Instead, the states are referenced in an additional column (Thermodynamic state). Additions use this column in cases where the state is not available in the subtitle line.

Additionally, modified the parsing/util.py script to use 'rt' mode in the case of .gz files. Alternatively, could have switched the input to use 'rt' mode but think this would effect the backflips for bz2 at the beginning of util.py.

Hope to receive suggestions to improve these modifications.

### DIFF
--- a/src/alchemlyb/tests/parsing/test_gmx.py
+++ b/src/alchemlyb/tests/parsing/test_gmx.py
@@ -4,8 +4,6 @@
 
 from alchemlyb.parsing.gmx import extract_dHdl, extract_u_nk
 from alchemtest.gmx import load_benzene
-from alchemtest.gmx import load_expanded_ensemble_case_1
-from alchemlyb.parsing.util import anyopen
 
 
 def test_dHdl():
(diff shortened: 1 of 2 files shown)


src/alchemlyb/parsing/util.py
```diff
@@ -47,7 +47,10 @@ def anyopen(filename, mode='r'):
     ext = os.path.splitext(filename)[1]
 
     if ext in extensions:
-       opener= extensions[ext]
```
orbeckst: This appears to have been a bug, right?

I'd prefer this to be a separate PR so that each PR addresses one specific problem and not a hodgepodge of things. Open an issue that identifies the problem and submit a PR + test for this issue. (Remove it from *this* PR).

We can discuss writing the test on the new PR.

src/alchemlyb/parsing/gmx.py
```diff
@@ -148,6 +170,18 @@ def _extract_state(xvg):
                 statevec = eval(line.strip().split(' = ')[-1].strip('"'))
                 break
 
+    if state is None:
```
orbeckst: add comment that you're parsing xe  data

orbeckst:
Now that  PR #45 is merged, please rebase your branch against master (to get rid of the gz_open code introduced here). The `push -f` to this branch with the re-written history. 

Add tests to make codecov happy...

PR #45, opened 2017-12-04
issue-44-parser-anyopen-gzip 

trje3733:
When parsing .gz files a TypeError occurs. This is because 'r' mode reads in binary mode. Changed util.py to handle gzip similar to bz2.

### DIFF
--- a/src/alchemlyb/parsing/util.py
+++ b/src/alchemlyb/parsing/util.py
@@ -9,7 +9,7 @@ import zipfile
 # need to do some backflips to support Python 2 bz2 behavior
 # bz2 in Python 2 doesn't have an open function, and in Python 3
 # the BZ2File class only does binary mode
-# need to do some backflips to support Python 2 gzip behavior
+# similar changes need to be made for gzip
 # gzip in Python 2 doesn't have a text mode, and in Python 3
 # gzip does have text mode
 try:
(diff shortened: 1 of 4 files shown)


src/alchemlyb/parsing/util.py
```diff
@@ -9,15 +9,23 @@
...
 # bz2 in Python 2 doesn't have an open function, and in Python 3
 # the BZ2File class only does binary mode
+# similar changes need to be made for gzip
+# gzip in Python 2 doesn't have a text mode, and in Python 3
+# gzip does have text mode
 try:
     bz2.open
 except AttributeError:
     bz2_open = bz2.BZ2File
+    gzip_open = gzip.open
```
orbeckst: Please put this in a separate try/except statement that does not depend on bz2. Decouple everything. Find a different way to test for gz specifically. `six` also has ways to tell you if it is Py 2 or Py 3 – use those if nothing else works.

orbeckst:
The codecov feedback should also show green.

trje3733:
@orbeckst I decoupled the checks at the beginning of util.py and added a test to ensure .gz lines are read in text mode. Let me know if you have any suggestions for changes.

orbeckst:
Thanks!

PR #58, opened 2018-10-15
Correct TI error estimation

harlor:
The error estimation for the free energy differences doesn't consider covariances. So I suggest to also take them into account.

### DIFF
--- a/src/alchemlyb/estimators/__init__.py
+++ b/src/alchemlyb/estimators/__init__.py
@@ -1,2 +1,3 @@
 from .mbar_ import MBAR
 from .ti_ import TI
+from .ti_cubic_ import TICUBIC
--- a/src/alchemlyb/estimators/ti_.py
+++ b/src/alchemlyb/estimators/ti_.py
@@ -54,12 +54,14 @@ class TI(BaseEstimator):
         means = dHdl.mean(level=dHdl.index.names[1:])
         variances = np.square(dHdl.sem(level=dHdl.index.names[1:]))
         
+        # get the lambda names
+        l_types = dHdl.index.names[1:]
+
         # obtain vector of delta lambdas between each state
         dl = means.reset_index()[means.index.names[:]].diff().iloc[1:].values
 
         # apply trapezoid rule to obtain DF between each adjacent state
         deltas = (dl * (means.iloc[:-1].values + means.iloc[1:].values)/2).sum(axis=1)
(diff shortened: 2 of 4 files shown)


harlor:
> Sampling at independent lambdas should be uncorrelated, since they are expected to come from independent simulations. Nonzero covariance is likely to be noise.

I completely agree. I was a bit imprecise. Acually I don't mean covariances between the mean dhdls but of the trapezoid areas. When one mean dhdl changes it changes the area of 2 trapezoids (Except it is a boundary value). 

Example with 3 dhdl-mean-values - 3 lambdas - 2 delta_lamdas:

Integral calculates as:
I = deltal_1*(y1+y2)/2 + deltal_2*(y2+y3)/2

Using propagation of uncertainty the integral has the variance:
dI^2 = (dy1*deltal_1/2)^2 + (dy2*(deltal_1+deltal_2)/2)^2 + (dy3*deltal_3/2)^2

mrshirts:
But we can rewrite the contribution of the trapezoids into a sum of independent calculations  I.e. although it's _derived_ as the area of a set of trapezoids, it is writeable (give it a try!) as:

\sum w_i <dh/dl>_i  

where w_i is a weight determined by the choice of lambdas and <dh/dl>_i  is the ensemble average of dh/dl at state i. No correlations are necessary; at best, they should cancel out.

Note that I realized that I actually haven't looked at THIS code in alchemlyb.  The error analysis code I wrote is at: https://github.com/MobleyLab/alchemical-analysis.  If the error analysis calculation in alchemlyb does not reflect the code in alchemical-analysis, then the alchemlyb should be checked and updated.   

But it is possible (and far preferable) to do WITHOUT any correlation matrix. Any autocorrelation matrix approach assume simple error propagation, which may not quite hold, and thus even if the overall effect is correct in the limit that simple error propagation is correct, they will introduce some error for most calculations.

harlor:
> Note that I realized that I actually haven't looked at THIS code in alchemlyb. The error analysis code I wrote is at: https://github.com/MobleyLab/alchemical-analysis. If the error analysis calculation in alchemlyb does not reflect the code in alchemical-analysis, then the alchemlyb should be checked and updated.

If I'm not mistaken, the error estimates differ. With this pull request I actually try to reproduce the results from alchemical-analysis.

harlor:
Sorry these commits should have never appeared here. This PR can be closed.

Issue #28, opened 2017-05-16
BAR estimator


ianmkenney:
Create a BAR estimator that:

1. takes `u_nk` as an argument
2. computes the free energy differences between each state along with their uncertainties

dotsdl:
@ianmkenney can you open a PR with your current BAR implementation? I know it wasn't working quite right yet, but that would allow more eyes to have a look (namely, mine :D).

orbeckst:
@harlor I am assigning this to you as you've been moving PR #60 forward.

harlor:
BTW - Besides a review is there something else missing in the PR?

PR #60, opened 2018-10-17
BAR Estimator

harlor:
This version seems to reproduce results i get using alchemical-analysis.


mrshirts:
Can you explain a little more what is different?  Also, be specific as to what tests you did to compare the alchemical-analysis and alchemlyb 

Note that the error estimate generated by BAR ARE correlated and so it's not a great idea to compute the overall uncertainty by adding them in the square.  Correlations are very hard to compute between neighboring BAR calculations.  Bootstrapping the entire data set is almost certainly the best way to do this.

harlor:
I used different input files to test this (As explained in https://github.com/alchemistry/alchemlyb/pull/61#issuecomment-430784175)

Indeed - unfortunately I am currently just calculating the sum of the squares. I guess bootstrapping just requires to have a sufficient amount of uncorrelated samples in u_kln? As suggested in https://github.com/choderalab/pymbar/issues/304#issuecomment-408980021 do you think 10 samples are sufficient to do an error estimation? Of course this is something that should be adjustable but to have a default value for this would still be nice.

dotsdl:
@harlor, thanks for this contribution! We've been wanting a BAR implementation for some time (#28), so this will be great to finally land. I'll have time to review sometime this weekend.

harlor:
Thank you @dotsdl!

@mrshirts 
> It probably makes sense to do bootstrapping as a general utility (perhaps inside alchemlyb, as perhaps a wrapper for all methods?), since it's useful for ALL calculations, not just BAR.

ok, maybe we can then just return nan-values for unknown uncertainties in the bar esimator here? Returning uncertainties that are known to be underestimated is not a good idea I guess.

dotsdl:
@harlor to move this one forward, we'll need some tests added to [test_fep_estimators.py](https://github.com/alchemistry/alchemlyb/blob/master/src/alchemlyb/tests/test_fep_estimators.py). You can create a class `TestBAR` similar to `TestMBAR` that tests the results from this estimator, which at the moment is an almost one-liner. Feel free to add additional tests as needed to help define the correct functionality of the `BAR` estimator.

harlor:
@dotsdl I added tests for the BAR estimator's value and its uncertainties - I also added tests for the uncertainty of the MBAR estimator and compared the results with [alchemical-analysis](https://github.com/MobleyLab/alchemical-analysis) where I find exact same values by using:
`alchemical_analysis.py -g -p lambda_ -t 300 -i 0 -u kBT -r 6`

@mrshirts Instead of calculating underestimated uncertainties, the estimator now just gives NaN for unknown uncertainties.

orbeckst:
The `predict()` method only contains `pass` and [is not tested](https://codecov.io/gh/alchemistry/alchemlyb/pull/60/diff). Can't we just drop it? Or does it have to be there – @dotsdl ?

src/alchemlyb/estimators/bar_.py
```diff
@@ -0,0 +1,135 @@
...
+                                     index=self.states_)
+
+        # yield standard deviation d_delta_f_ between each state
+        self.d_delta_f_ = pd.DataFrame(np.sqrt(ad_delta + ad_delta.T),
+                                       columns=self.states_,
+                                       index=self.states_)
+
+        return self
+
+    def predict(self, u_ln):
```
orbeckst: Can we drop it?

(It's neither documented nor tested.)

src/alchemlyb/estimators/bar_.py
```diff
@@ -0,0 +1,135 @@
...
+
+                # See https://github.com/alchemistry/alchemlyb/pull/60#issuecomment-430720742
+                # Error estimate generated by BAR ARE correlated
+
+                # Use the BAR uncertainties between two neighbour states
+                if j == 0:
+                    dout.append(d_deltas[i:i + j + 1].sum())
+                # Other uncertainties are unknown at this point
+                else:
+                    dout.append(float('NaN'))
```
orbeckst: Or use `np.NaN` directly, given that we already use numpy. Not sure what's more Pythonic although [float("nan") seems to be the way to go without numpy](https://stackoverflow.com/a/19374296).

Bogus performance "argument" in favor of numpy:
```python
In [17]: %time float("NaN")
CPU times: user 4 µs, sys: 0 ns, total: 4 µs
Wall time: 9.06 µs
Out[17]: nan

In [18]: %time np.NaN
CPU times: user 4 µs, sys: 0 ns, total: 4 µs
Wall time: 7.87 µs
Out[18]: nan
```

src/alchemlyb/estimators/bar_.py
orbeckst: If you keep then there should be a test that shows that it does nothing.

harlor: This is comes from the implementation of MBAR which I used as a starting point. I don't find any usages of this method in the project so let's drop it...

orbeckst:
Looks good to me from the pure code/api side of things.

Quick approval from @mrshirts (and/or @davidlmobley ) regarding the veracity of the implementation would be highly appreciated! – Essentially, are you satisfied with the evidence that @harlor has shown to demonstrate that this implementation produces correct answers (i.e., same answer as `alchemical-analysis`)?

src/alchemlyb/estimators/bar_.py
orbeckst: If it's in the MBAR class then we should also remove it from there (in a different PR).

orbeckst:
Ping @mrshirts @davidlmobley .

orbeckst:
I haven't heard anything against merging this PR so I am now merging it. If anything else comes up later, please open a new issue.

Thanks @harlor !

davidlmobley:
Yes, I think this is fine. Sorry for the delay; huge series of deadlines. 

@hbaumann do you have a set of data from equilibrium free energy calculations you can run through this to doublecheck results seem correct?

Issue #59, opened 2018-10-15
Gromacs XVG parser - Calculation of the reduced potential goes badly wrong 


harlor:
Assuming to find values of U and pV in the second and last column can go badly wrong if they are not given. 

https://github.com/alchemistry/alchemlyb/blob/e5b2e581528769a205cab17dfae7d1b9b2f8f2f5/src/alchemlyb/parsing/gmx.py#L49-L61

What do you think: Should we print a warning when these columns are not given and continue with what is given or directly throw an exception?

Btw: Is it really necessary to add pV if the simulations are in NVT? Why do we add U?

mrshirts:
If the simulations are carried out at 2 different temperatures, we need U, as the ratio of Boltzmann factors involved the total energy, not just the difference.

Certainly, if they are not given, this should be caught well before. 

For NVT, one shouldn't need PV.  I'll need to check to see how this was handled in alchemical-analysis.

dotsdl:
Hi @harlor, welcome to `alchemlyb`! We're always looking for ways to improve the existing components of the library. Do you have an example dataset using NVT? It would be worthwhile to generate one, perhaps with a single molecule such as [benzene](https://github.com/alchemistry/alchemtest/blob/master/src/alchemtest/gmx/benzene/descr.rst), but in NVT instead of NPT. @ianmkenney, do you have a way to do this by chance?

harlor:
I have NPT and NVT datasets available but all without U.

If I understand the alchemical-analysis parser correctly it ads pV if its available and does not add U.

Maybe if nothing else os specified one could add pV if its available and add U if its available?

PR #62, opened 2018-10-21
Add pV and U to reduced potential when they are given

harlor:
As explained in https://github.com/alchemistry/alchemlyb/issues/59 The current parser assumes U and pV to be given in the second respectively last column - which is not very robust.

With these changes the parser adds them to the potential differences only when they are given. 

Moreover I am interested if these changes pass the tests.

### DIFF
--- a/src/alchemlyb/parsing/gmx.py
+++ b/src/alchemlyb/parsing/gmx.py
@@ -29,7 +29,9 @@ def extract_u_nk(xvg, T):
 
     """
 
-    col_match = r"\xD\f{}H \xl\f{}"
+    h_col_match = r"\xD\f{}H \xl\f{}"
+    pv_col_match = 'pV'
+    u_col_match = 'Total Energy'
     beta = 1/(k_b * T)
 
     state, lambdas, statevec = _extract_state(xvg)
@@ -43,22 +45,31 @@ def extract_u_nk(xvg, T):
     times = df[df.columns[0]]
 
     # want to grab only dH columns
-    DHcols = [col for col in df.columns if (col_match in col)]
+    DHcols = [col for col in df.columns if (h_col_match in col)]
     dH = df[DHcols]
 
-    # not entirely sure if we need to get potentials relative to
-    # the state actually sampled, but perhaps needed to stack
-    # samples from all states?
-    U = df[df.columns[1]]
-
     # gromacs also gives us pV directly; need this for reduced potential
-    pV = df[df.columns[-1]]
+    pv_cols = [col for col in df.columns if (pv_col_match in col)]
+    pv = None
+    if pv_cols:
+        pv = df[pv_cols[0]]
+
+    # gromacs also gives us total energy U directly; need this for reduced potential
+    u_cols = [col for col in df.columns if (u_col_match in col)]
+    u = None
+    if u_cols:
+        u = df[u_cols[0]]
 
     u_k = dict()
-    cols= list()
+    cols = list()
     for col in dH:
         u_col = eval(col.split('to')[1])
-        u_k[u_col] = beta * (dH[col].values + U.values + pV.values)
+        # calculate reduced potential u_k = dH + pV + U
+        u_k[u_col] = beta * dH[col].values
+        if pv_cols:
+            u_k[u_col] += beta * pv.values
+        if u_cols:
+            u_k[u_col] += beta * u.values
         cols.append(u_col)
 
     u_k = pd.DataFrame(u_k, columns=cols,
--- a/src/alchemlyb/parsing/gmx.py
+++ b/src/alchemlyb/parsing/gmx.py
@@ -31,7 +31,7 @@ def extract_u_nk(xvg, T):
 
     h_col_match = r"\xD\f{}H \xl\f{}"
     pv_col_match = 'pV'
-    u_col_match = 'Total Energy'
+    u_col_match = ['Total Energy', 'Potential Energy']
     beta = 1/(k_b * T)
 
     state, lambdas, statevec = _extract_state(xvg)
@@ -54,8 +54,8 @@ def extract_u_nk(xvg, T):
     if pv_cols:
         pv = df[pv_cols[0]]
 
-    # gromacs also gives us total energy U directly; need this for reduced potential
-    u_cols = [col for col in df.columns if (u_col_match in col)]
+    # gromacs also gives us total/potential energy U directly; need this for reduced potential
+    u_cols = [col for col in df.columns if any(single_u_col_match in col for single_u_col_match in u_col_match)]
     u = None
     if u_cols:
         u = df[u_cols[0]]
--- a/src/alchemlyb/tests/parsing/test_gmx.py
+++ b/src/alchemlyb/tests/parsing/test_gmx.py
@@ -5,6 +5,9 @@
 from alchemlyb.parsing.gmx import extract_dHdl, extract_u_nk
 from alchemtest.gmx import load_benzene
 from alchemtest.gmx import load_expanded_ensemble_case_1, load_expanded_ensemble_case_2, load_expanded_ensemble_case_3
+from alchemtest.gmx import load_water_particle_with_total_energy
+from alchemtest.gmx import load_water_particle_with_potential_energy
+from alchemtest.gmx import load_water_particle_without_energy
 
 
 def test_dHdl():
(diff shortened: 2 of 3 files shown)


harlor:
> This looks sensible. However, there should also be a test case without either/or pV and U columns. 
You might be able to edit the benzene files or add new datasets. Either way, open a PR in https://github.com/alchemistry/alchemtest/ to add the data and then add tests here, please.

Agree - when I find time to work my way into the testing system I will do that!

orbeckst:
I think you are right regarding our tests having been wron regarding parsing (or lack thereof) of the correct columns. However, who does only statistical inefficiency change and not the actual free energies? (Sorry, I'm typing from mobile and it is difficult to look at code to answer my question myself). 

@dldotson @ianmkenney can you you also have a quick look and confirm?

orbeckst:
Er, Ping @dotsdl --- sorry, mistyped your handle in previous comment.

src/alchemlyb/parsing/gmx.py
```diff
@@ -29,7 +29,9 @@ def extract_u_nk(xvg, T):
 
     """
 
-    col_match = r"\xD\f{}H \xl\f{}"
+    h_col_match = r"\xD\f{}H \xl\f{}"
+    pv_col_match = 'pV'
+    u_col_match = 'Total Energy'
```
dotsdl: Checking the files you provided @harlor (https://userpage.fu-berlin.de/dominikwille/nvt300K.tgz), I see "Potential Energy (kJ/mol)" in the `ti_0.xvg` file. Should we also try and match that?

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -100,7 +100,7 @@ def slicer(self, *args, **kwargs):
                                  (True, gmx_benzene_dHdl(), 2001),  # 0.00:  g = 1.0559445620585415
                                  (True, gmx_benzene_u_nk(), 2001),  # 'fep': g = 1.0560203916559594
                                  (False, gmx_benzene_dHdl(), 3789),
-                                 (False, gmx_benzene_u_nk(), 3789),
+                                 (False, gmx_benzene_u_nk(), 3571),
```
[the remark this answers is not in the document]
dotsdl: Makes sense to me. Also, I believe the splitting resulting from the statistical inefficiency can vary a bit by machine it's executed on, so these are a bit tenuous at best. We mostly just want to know when the CI server sees a change from what is known to occur.

dotsdl:
@harlor, thanks for putting this together. I see the issue as well. The [Gromacs benzene dataset](https://github.com/alchemistry/alchemtest/blob/master/src/alchemtest/gmx/benzene/descr.rst) is indeed missing the potential energy (U_i) assumed by the `u_nk` parser, so currently it's giving really wrong results as in this [test](https://github.com/alchemistry/alchemlyb/blob/master/src/alchemlyb/tests/parsing/test_gmx.py#L23-L37).

I am very much in favor of these changes. I'm trying to think of how we should alert users to issues where the parser can still give meaningful results, but that the result may not be in absolute terms (such as when potential energy is missing).

Perhaps the best approach is to throw an exception when either the `pV` or `Potential Energy` column is missing? We can then have keyword arguments such as `use_pV=False` (default `True`) and `use_U=False` (default `True`) to make the parser proceed with ignoring these values, giving "reduced potentials" without them. We can do the same with `Total Energy` with a `use_H` (default `False`). 

Exceptions will be raised if one of these keyword args is given but can't be honored with the given dataset, or if two are mutually exclusive (such as `use_U=True` and `use_H=True`).

Thoughts? This doesn't have to be in the scope of this PR, but I think it's time we add some knobs to these parsers.

src/alchemlyb/parsing/gmx.py
harlor: @mrshirts explained that in https://github.com/alchemistry/alchemlyb/issues/59#issuecomment-430721922

I uploaded a dataset with total energy here: http://dominikwille.userpage.fu-berlin.de/nvt300K_u.tgz

dotsdl: Okay, but I *think* [depending on MDP options given to gromacs](https://alchemlyb.readthedocs.io/en/latest/parsing/alchemlyb.parsing.gmx.html) it's possible to have a dataset without "Total Energy" but with "Potential Energy". I was thinking that we could handle this in [another PR with kwargs](https://github.com/alchemistry/alchemlyb/pull/62#issuecomment-434935124), but I don't believe this should be neglected.

Could you add a check for the existence of a "Potential Energy" column as you did for the "Total Energy" column, and if present, use this instead of the "Total Energy" column? That would satisfy me for the purposes of this PR

dotsdl:
I think if we address the presence of "Potential Energy", we'll be good to go on a merge for this. 

@orbeckst, is there anything else you see? The benzene dataset doesn't have either "Total Energy" or "Potential Energy", so I think it satisfies [your earlier comment](https://github.com/alchemistry/alchemlyb/pull/62#pullrequestreview-168569475) already.

harlor:
I opened a PR in [alchemtest](https://github.com/alchemistry/alchemtest/pull/30) with datasets that allow to write a test for this PR.

@orbeckst Do you consider it sufficient to check a certain value of the reduced potential?

orbeckst:
I reviewed https://github.com/alchemistry/alchemtest/pull/30.

For tests I would generally use aggregate calculations such as a mean or a sum, in addition to specific values.

harlor:
Ok I added tests for specific elements and for the sum of elements on the diagonal.

Btw: Is it possible/a good idea to use hash values of dataframes to write tests?

orbeckst:
> Btw: Is it possible/a good idea to use hash values of dataframes to write tests?

Not a good idea if it contains floating point numbers.

For those you can never test for equality. Instead you do
```python
from numpy.testing import assert_almost_equal

assert_almost_equal(calculated, reference)
```

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +106,65 @@ def test_dHdl_case3():
...
+    """
+
+    # Load dataset
+    dataset = load_water_particle_with_potential_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert _diag_sum(dataset) == 16674041445589.646
+
+    # Check one specific value in the dataframe
+    assert float(extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0]) == -15659.655560881085
```
orbeckst: needs to use `assert_almost_equal` or similar. You cannot compare floats for equality because this depends on hardware etc.

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +106,65 @@ def test_dHdl_case3():
...
+    """
+
+    # Load dataset
+    dataset = load_water_particle_without_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert _diag_sum(dataset) == 20572988148877.555
+
+    # Check one specific value in the dataframe
+    assert float(extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0]) == 18.134225023007403
```
orbeckst: float comparison!

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +106,65 @@ def test_dHdl_case3():
...
+def test_u_nk_without_energy():
+    """Test that the reduced potential is calculated correctly when no energy is given.
+
+    """
+
+    # Load dataset
+    dataset = load_water_particle_without_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert _diag_sum(dataset) == 20572988148877.555
```
orbeckst: float comparison!

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +106,65 @@ def test_dHdl_case3():
...
+def test_u_nk_with_potential_energy():
+    """Test that the reduced potential is calculated correctly when the potential energy is given.
+
+    """
+
+    # Load dataset
+    dataset = load_water_particle_with_potential_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert _diag_sum(dataset) == 16674041445589.646
```
orbeckst: float comparison!

harlor:
As you suggested I use `assert_almost_equal` now.

>> Btw: Is it possible/a good idea to use hash values of dataframes to write tests?

> Not a good idea if it contains floating point numbers.

Good to know!

orbeckst:
The tests fail because the accuracy is not high enough. Add
```python
assert_almost_equal(..., decimal=6)
```
to decrease accuracy.

When you work with continuous integration, check the output of the tests and if it failed (red X), fix it/ask for advice for how to fix.

You should also be able to run the tests locally with
```
pytest
```
to catch anything obvious so that you don't have to wait for Travis.

orbeckst:
P.S.: I think it was `decimal`, could also be `decimals`.

orbeckst:
P.P.S.: Also merge master into your branch.

harlor:
I will adopt local testing into my development workflow!

It turned out that I get the datasets from alchemtest in a different order on my machine - which causes not only these tests but also `TestStatisticalInefficiency` to give different results.

Is using something like `sorted(glob(...))` instead of `glob(...)` in alchemtest, which sorts the datasets by filename a good idea?

See: https://stackoverflow.com/questions/6773584/how-is-pythons-glob-glob-ordered

orbeckst:
yes!

(I just merged your PR https://github.com/alchemistry/alchemtest/pull/31)

harlor:
Thanks - is this PR also ready then?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +107,77 @@ def test_dHdl_case3():
...
+
+    # Load dataset
+    dataset = load_water_particle_with_total_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert_almost_equal(_diag_sum(dataset), 47611377946.58586, decimal=4)
+
+    # Check one specific value in the dataframe
+    assert_almost_equal(
+        float(extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0]),
```
orbeckst: Why is `float` needed?

orbeckst: Should probably work without.

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +107,77 @@ def test_dHdl_case3():
...
+
+    # Load dataset
+    dataset = load_water_particle_with_potential_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert_almost_equal(_diag_sum(dataset), 16674041445589.646, decimal=2)
+
+    # Check one specific value in the dataframe
+    assert_almost_equal(
+        float(extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0]),
```
orbeckst: Why is `float` needed?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +107,77 @@ def test_dHdl_case3():
...
+
+    # Load dataset
+    dataset = load_water_particle_without_energy()
+
+    # Check if the sum of values on the diagonal has the correct value
+    assert_almost_equal(_diag_sum(dataset), 20572988148877.555, decimal=2)
+
+    # Check one specific value in the dataframe
+    assert_almost_equal(
+        float(extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0]),
```
orbeckst: Why is `float` needed?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -103,3 +107,77 @@ def test_dHdl_case3():
...
+
+    for leg in dataset['data']:
+        for filename in dataset['data'][leg]:
+            u_nk = extract_u_nk(filename, T=300)
+
+            # Calculate the sum of diagonal elements:
+            for i in range(len(dataset['data'][leg])):
+                ds += u_nk.iloc[i][i]
+
+    return ds
```
orbeckst: add newline

orbeckst:
@dotsdl you will also need to check and approve. Apparently, we have strict merge rules here... not enough to only satisfy one reviewer ;-).

orbeckst:
@dotsdl do you want to review or should I go ahead and merge? If I don't hear anything by 6pm AZ I'll merge using admin powers.

dotsdl:
@orbeckst I should be able to review this during the weekend. If I don't manage to by Monday 11/26, please proceed with the merge.

harlor:
I don't know why the build of the tests failed - it worked after a restart.

dotsdl:
It appeared to be a problem with building scikit-learn wheels, so something upstream was broken. Looks resolved now. I'll be reviewing this later today. Thanks @harlor!

dotsdl:
Looks good to me @harlor! Thanks for all the work you put into on this. And thanks @orbeckst for moving this along! Merging!

PR #65, opened 2018-11-03
Set Travis CI to test against Python 3.7

dotsdl:
Requires us to use `dist: xenial` (Ubuntu 16.04) to add this Python version to the test matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: python
-sudo: false
+sudo: required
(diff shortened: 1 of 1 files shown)


dotsdl:
Thanks to @harlor's work in alchemistry/alchemtest/pull/31 and #69, this now passes the tests. Given that this serves only to add Python 3.7 coverage, I'm going to merge.

Issue #66, opened 2018-11-04
Switching Travis CI to Xenial (Ubuntu 16.04) gives different statistical inefficiency test results


dotsdl:
I recently tried changing our [Travis CI](https://github.com/alchemistry/alchemlyb/blob/master/.travis.yml) configuration to use `dist: xenial` so that we could add testing against Python 3.7. I've made this into a PR (#65), since it gave errors that will require some discussion to resolve.

From what I can see, we get the following failures:
1. It looks like the change in underlying Ubuntu version gives different results for statistical inefficiency slicing.
2. For Python 3.7, the Amber parsers throw exceptions related to their use of generators.

We'll need to resolve these in the PR before merging to master.

harlor:
For 2. see https://github.com/alchemistry/alchemlyb/pull/69

dotsdl:
Resolved with #65.

PR #69, opened 2018-11-22
Make amber parser compatible with python 3.7

harlor:
As explained in https://github.com/alchemistry/alchemlyb/issues/66 the amber parser does not work in Python 3.7.

These changes should fix this.

See https://www.python.org/dev/peps/pep-0479/#sub-proposal-decorator-to-explicitly-request-current-behaviour


dotsdl:
Merging!

PR #74, opened 2019-04-14
Amber parsers use kT units

### LINKED ISSUE #56: Amber FEP parser should return reduced potentials u_nk
The current version of the Amber FEP parser returns energies (as used to be the case in alchemical-analysis). alchemlyb prescribes that the [u_nk standard form of the data be in *reduced units*](http://alchemlyb.readthedocs.io/en/latest/parsing.html#u-nk-standard-form). Specifically, this means that
- the pV term is added
- the units are in kT (i.e., multiply with beta = 1/kT)
See https://github.com/alchemistry/alchemlyb/blob/e5b2e581528769a205cab17dfae7d1b9b2f8f2f5/src/alchemlyb/parsing/gmx.py#L61

(All  energies are given in [units of kT](http://alchemlyb.readthedocs.io/en/latest/parsing.html#a-note-on-units).)

- [ ] [`amber.extract_u_nk()`](http://alchemlyb.readthedocs.io/en/latest/parsing/alchemlyb.parsing.amber.html#alchemlyb.parsing.amber.extract_u_nk) needs additional temperature `T` argument
- [ ] make raw data into reduced "u_nk" potentials 
- [ ] adjust tests

### LINKED ISSUE #57: Amber TI parser should provide dimensionless dH/dl
The current implementation of the Amber TI parser ([amber.extract_dHdl](http://alchemlyb.readthedocs.io/en/latest/parsing/alchemlyb.parsing.amber.html#alchemlyb.parsing.amber.extract_dHdl)) returns a dataframe with native Amber units (as used to be the case in alchemical-analysis) instead of the [prescribed reduced dHdl standard form](http://alchemlyb.readthedocs.io/en/latest/parsing.html#dhdl-standard-form): the potential energy derivatives should be divided by 1/kT.

See for example https://github.com/alchemistry/alchemlyb/blob/e5b2e581528769a205cab17dfae7d1b9b2f8f2f5/src/alchemlyb/parsing/gmx.py#L132

Todo:
- [ ] `amber.extract_dHdl()` needs additional temperature `T` argument
- [ ] make data dimensionless (something like `dHdl = beta * dHdL` with `beta = 1/(kB*T)`
- [ ] adjust tests

(See also #56 )

orbeckst:
- fix #57 (TI parser)
- fix #56 (FEP parser)

### DIFF
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -18,6 +18,12 @@ from .util import anyopen
 
 logger = logging.getLogger("alchemlyb.parsers.Amber")
 
+# TODO: perhaps move constants elsewhere?
+# these are the units we need for dealing with Amber, which uses
+# kcal/mol for energies  http://ambermd.org/Questions/units.html
+# (kB in kcal/molK)
+k_b = 1.9872041e-3
+
 
 def convert_to_pandas(file_datum):
     """Convert the data structure from numpy to pandas format"""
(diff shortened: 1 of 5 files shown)


orbeckst:
TODO: Need to add tests for Amber/BAR and Amber/MBAR.

orbeckst:
Thanks for the quick review. 


> One thing that I stumbed across when testing is that pip installing alchemtest does not contain the amber testing data and therefore must be installed from git.

As soon as we have a proper 0.2.0 release of the testdata (and the library itself) will the Amber files be installed. They are not yet in the released version.

PR #76, opened 2019-04-14
make NAMD parser return energies in kT

### LINKED ISSUE #75: NAMD parser does not take temperature as argument
The NAMD parser #7 (for FEP) does not take `T` as an argument and does not return `u_nk` in proper units of kT.

orbeckst:
fix #75 

- move NAMD FEP test from parsing to test_fep
- take T as argument

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,5 +29,5 @@ setup(name='alchemlyb',
       license='BSD',
       long_description=open('README.rst').read(),
       tests_require = ['pytest', 'alchemtest'],
-      install_requires=['numpy', 'pandas', 'pymbar', 'scipy', 'scikit-learn']
+      install_requires=['numpy', 'pandas>=0.23.0', 'pymbar', 'scipy', 'scikit-learn']
       )
--- a/src/alchemlyb/parsing/namd.py
+++ b/src/alchemlyb/parsing/namd.py
@@ -30,69 +30,67 @@ def extract_u_nk(fep_file, T):
     """
     beta = 1/(k_b * T)
 
-    # lists to get timesteps and work values of each window
+    # lists to get times and work values of each window
     win_ts = []
     win_de = []
 
     # create dataframe for results
-    u_nk = pd.DataFrame(columns=['timestep','fep-lambda'])
+    u_nk = pd.DataFrame(columns=['time','fep-lambda'])
 
     # boolean flag to parse data after equil time
     parsing = False
 
     # open and get data from fep file.
     with anyopen(fep_file, 'r') as f:
-        data = f.readlines()
+        for line in f:
+            l = line.strip().split()
 
-    for line in data:
-        l = line.strip().split()
+            # this line marks end of window; dump data into dataframe
+            if '#Free' in l:
 
-        # this line marks end of window; dump data into dataframe
-        if '#Free' in l:
+                # convert last window's work and times values to np arrays
+                win_de_arr = beta * np.asarray(win_de)
+                win_ts_arr = np.asarray(win_ts)
 
-            # convert last window's work and timestep values to np arrays
-            win_de_arr = beta * np.asarray(win_de)
-            win_ts_arr = np.asarray(win_ts)
+                # extract lambda values for finished window
+                # lambda1 = sampling lambda (row), lambda2 = evaluated lambda (col)
+                lambda1 = "{0:.2f}".format(float(l[7]))
+                lambda2 = "{0:.2f}".format(float(l[8]))
 
-            # extract lambda values for finished window
-            # lambda1 = sampling lambda (row), lambda2 = evaluated lambda (col)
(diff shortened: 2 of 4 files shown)


orbeckst:
Note: tests do not pass yet, input welcome!

orbeckst:
See also #56 and PR #74 .

orbeckst:
Discussion regarding this PR is mainly at issue #75.

dotsdl:
Scheduled for review on 2019.05.05. Thanks @orbeckst! Apologies for the delay.

orbeckst:
@dotsdl the discussion in #75 is important because there are questions about what the correct way to proceed is.

Please do look at the code, that would be great (maybe I messed up somewhere), but also look at the scientific discussion for context.

orbeckst:
Btw – don't apologize. Thank you for making time!!!

dotsdl:
Thanks @orbeckst, I am beginning my review. Trying a new approach where I specifically time-block my open source efforts (by PR or issue). This works well for other spheres of my life, so thought I'd go for it here. Keeps me from getting overwhelmed by the "what do I work on?" question, since that can spin for a while before anything actually happens.

dotsdl:
Made some adjustments. They are as follows:

1. Changed 'timestep' column to 'time', for consistency with the [u_nk standard form](https://alchemlyb.readthedocs.io/en/latest/parsing.html#u-nk-standard-form).
2. Removed call to `file.readlines()` and instead iterate through file lines as a stream.
3. Missing values are left as NaNs instead of converting to zeros. Zeros indicate no energy difference, while NaNs indicate no measure was taken. Downstream estimators should be written to handle missing values gracefully.

I still don't know how to account for the .9 kT discrepancy we observe in the test. When I take the dataframe produced from combining the forward and backward data and multiply it by `kT` for NAMD (effectively *just* changing the units by a 0.6 scale factor), I get what the test expects for the resulting free energy when fitting BAR. Somehow changing units by a scale factor substantially changes the result BAR gives.

@mrshirts, is there anything in the `pymbar.BAR` code that might give us a different result if all inputs are rescaled by 1.7 (1/(kT) for NAMD with T=300)? That seems so odd to me. I hate to pass the buck, but the `alchemlyb.estimators.bar_.BAR.fit` method only does data gymnastics, and leaves the free energy calculation itself to the `pymbar.BAR` class.

Other than this (major) issue, @orbeckst this PR looks good. This is the correct approach for maintaining parsing consistency within `alchemlyb`, but I'd like us to hold on merging until we at least have a clue on how the result has changed. Sound good?

orbeckst:
When I set kT = 1 kcal/mol
```python
u_nk1 = namd.extract_u_nk(dataset['data']['forward'][0], T=300/0.6)
```
and
```python
kT_NAMD = 300/6 * namd.k_b
```
then (unsurprisingly) I get more or less the original 11.something.

But with T=300 I get the 0.9 disagreement.

@dotsdl what exactly did you do here:
> I still don't know how to account for the .9 kT discrepancy we observe in the test. When I take the dataframe produced from combining the forward and backward data and multiply it by kT for NAMD (effectively just changing the units by a 0.6 scale factor), I get what the test expects for the resulting free energy when fitting BAR. Somehow changing units by a scale factor substantially changes the result BAR gives.

From https://github.com/alchemistry/alchemlyb/issues/75#issuecomment-490281944 I took away that in principle we can use anything for the factor for reducing the input potentials, as long as we use the inverse on the output. 

Not solved yet.

mrshirts:
I probably didn't explain that clearly enough.  MBAR and BAR all assume that the data was generated according to the Boltzmann distribution at some temperature. kT has to be consistent with the Boltzmann ensemble that the data was generated with.  If it was a Boltzmann ensemble generated at 50 K, but I used kT equivalent to 300 K, that will not give physical results.

orbeckst:
@mrshirts thanks for comment https://github.com/alchemistry/alchemlyb/pull/76#issuecomment-490316958. This implies that having run the previous code directly on the input values in kcal/mol will have produced a systematically wrong solution. Taken all together (see https://github.com/alchemistry/alchemlyb/issues/75#issuecomment-512598519) I feel reasonably confident to believe the corrected code in this PR. 

@dotsdl once CI passes could you please do a final review? Thanks!

(And then we can actually do a 0.2.0 release!)

orbeckst:
P.S. The CI https://travis-ci.org/alchemistry/alchemlyb/builds/560232933 passed 30 mins ago but GitHub hasn't caught up...

orbeckst:
@dotsdl could you put this PR on your calendar for a quick final review, please? Thanks.

dotsdl:
Added! Reviewing Sunday afternoon. Thanks @orbeckst!

dotsdl:
Haven't forgotten about this, but have had to move my review to Tuesday night. Apologies, @orbeckst.

PR #86, opened 2019-07-24
gmx parsers now drop duplicate columns

dotsdl:
Some of our datasets, such as [`alchemtest.gmx.benzene`](https://github.com/alchemistry/alchemtest/tree/master/src/alchemtest/gmx/benzene), feature duplicate columns in their data files. Previously this was handled after initial parsing with `pandas.read_csv`, but as of the latest `pandas` release this function disallows duplicate column names on entry.

We now mark duplicate column names before entry into `pandas.read_csv`, and we filter these duplicate columns from the resulting dataframe after this initial parsing.

PR #87, opened 2019-07-26
Removed redundant duplicate removal from gmx u_nk parser

dotsdl:
This is follow-up work from #86. We don't want redundancy in duplicate removal here if we don't expect it to be necessary, as this can result in silently dropping things unintentionally or make issues harder to debug.


orbeckst:
Thanks!

Weirdly, codecov worked fine here but not in PR #78 ...

PR #89, opened 2019-07-26
Faster ti fixtures 88

### LINKED ISSUE #88: speed-up fixture collection in tests
The "collect fixtures" stage of the TI pytest takes a long time because we run parsing code as part of the fixtures, e.g. https://github.com/alchemistry/alchemlyb/blob/29648ead7cf2212b4f16a0fd38baacd4c300fea2/src/alchemlyb/tests/test_ti_estimators.py#L101-L110 

It would be better if these parts of the fixtures were run _at test time_ instead of fixture collection as (1) the set up time would be faster (better for interactive testing) and (2) failures would be reported during test run time and not collection.

We should be able to do this by simply delegating the *running* of the parsing code such as `gmx_benzene_coul_dHdl()` to the test function.

This was already done for FEP, see https://github.com/alchemistry/alchemlyb/blob/29648ead7cf2212b4f16a0fd38baacd4c300fea2/src/alchemlyb/tests/test_fep_estimators.py#L120-L129 and https://github.com/alchemistry/alchemlyb/blob/29648ead7cf2212b4f16a0fd38baacd4c300fea2/src/alchemlyb/tests/test_fep_estimators.py#L131-L133

orbeckst:
- fix #88 
- made TI test code more similar in structure to FEP test code

### DIFF
--- a/src/alchemlyb/tests/test_fep_estimators.py
+++ b/src/alchemlyb/tests/test_fep_estimators.py
@@ -7,8 +7,7 @@ import numpy as np
 import pandas as pd
 
 from alchemlyb.parsing import gmx, amber, namd
-from alchemlyb.estimators import MBAR
-from alchemlyb.estimators import BAR
+from alchemlyb.estimators import MBAR, BAR
 import alchemtest.gmx
 import alchemtest.amber
 import alchemtest.namd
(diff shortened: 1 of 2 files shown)


dotsdl:
lgtm @orbeckst! Thanks for doing this; it definitely makes the tests do less work upfront so we can iterate more quickly in development. I'm merging.

PR #93, opened 2020-01-15
Changes in light of upstream pymbar 3.0.5

dotsdl:
We need an explicit `return_theta=True` call now, as before this was implicitly working.

Also switched to using a list for alchemlyb.estimators.MBAR.method, as this is in line with the stated docs for inputs to pymbar.MBAR (we were using a tuple).

We have set the install requirements in setupy.py to now have `pymbar==3.0.5` to ensure users are staying up to the latest version, as we intend to track pymbar's development carefully.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,22 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 
+
+01/14/2020 dotsdl
+
+* 0.3.1
+
+Enhancements
+
+Deprecations
+
+Fixes
+  - added explicit `return_theta=True` for call to pymbar.MBAR.getFreeEnergyDifferences,
+    as this was happening prior to changes in `pymbar` without the explicit call
+
+Changes
+
+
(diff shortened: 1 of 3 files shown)


dotsdl:
I am recommending that we cut a 0.3.1 point release immediately after this is merged, as this will fix current broken behavior with MBAR.

orbeckst:
Get out a patch level release!

dotsdl:
Changes to `pymbar.mbar.MBAR.getFreeEnergyDifferences` in 3.0.5 caused the change in the number of items in the tuple that is returned if `return_theta=False` from 3 items to 2 items. Previously in 3.0.3, `return_theta=False` would still return 3 items, but the last one (`Theta`) would be `None`.

There are some API changes in the works for `pymbar` coming in 4.0, so we'll likely switch over to that new API for our `MBAR` estimator to track `pymbar`'s development pathway.

dotsdl:
I'll push out a patch release this evening following the release guide on the wiki. Thanks @orbeckst!

Issue #80, opened 2019-05-08
Bootstrapping: should it be at the alchemlyb level, or pymbar level?


mrshirts:
Should bootstrapping be implemented at the alchemlyb level, or pymbar level?  For MBAR, it would be better at the pymbar level, since it can be easily encapsulated (user doesn't have to worry about it), and one can request either uncertainty estimate.  

For BAR over several states, then the bootstrapping needs to be done at the level ABOVE the BAR call, since we need to bootstrap all of the data simultaneously before feeding it into BAR.  Same for EXP applied to a string of states. 

Thoughts?

orbeckst:
If there's a common function that we can use for "everything" (and just plugin the estimator) then alchemlyb would be a good place, I think – along the lines that we only have to write and test the code once. 

I generally like building blocks that I can freely combine. Something like
```python
def bootstrapped(data, estimator):
     ...
     return mean, error
```

Alternatively, we could hide the machinery in the `Estimator.fit()` function: enable it with `fit(..., ..., bootstrap=True)` and just do under the hood whatever needs to be done and set the appropriate output attributes. In this case, one might just use a pymbar bootstrap directly if available or roll our own for other estimators.

One advantage of doing it at the alchemlyb level is that it might not be difficult to run alchemlyb with dask (essentially, use the dask.DataFrame) and then the bootstrapping can be parallelized without effort. A while ago @dotsdl played around with alchemlyb and dask – I can't quite remember how much would need to be changed.

mrshirts:
> def bootstrapped(data, estimator):

That might be a good idea.  It could return a dictionary of all the bootstrapped results, along with the uncertainty estimate.   I'll think about how to organize this.  

One issue is that the data will look different with each estimator, thus requiring fairly different conditionals inside the bootstrapped data.  Also, if one was analyzing K states, calculating the free energy with BAR executed pairwise, one would want to bootstrap over the entire data set of K states; i.e. you would need to bootstrap the entire procedure, not over a single estimator.

dotsdl:
@mrshirts do you have a paper or writeup you can point to for this approach? I'd be happy to prototype something. We may be able to steal design inspiration from `scikit-learn` such as their `GridSearchCV` and `Pipeline` objects.

mrshirts:
So, I don't really have a good simple paper. http://www.alchemistry.org/wiki/Analyzing_Simulation_Results#Bootstrap_Sampling is a good summary.   

[...]

PR #94, opened 2020-01-20
WIP: bootstrapping submodule, parameter for estimators

dotsdl:
This PR addresses #80.

We have added an `alchemlyb.preprocessing.bootstrapping` submodule. The functions exposed by this module should work on both `dHdl` and `u_nk` standard forms, with a similar usage feel to the `alchemlyb.preprocessing.subsampling` functions. The following are needed to complete this PR:
- [ ] tests
- [ ] usage example(s)
- [ ] preprocessing documentation for `bootstrapping` submodule
- [ ] addition to base estimator class
- [ ] at least one block bootstrapping implementation

In addition to the function(s), we additionally want to add a bootstrapping keyword (or set of keywords) to each estimator's constructor, which changes how uncertainties are estimated internally to the estimator at the expense of more compute. This will be the approach we will push users toward using for simplicity, but we still want to expose the building blocks by way of the`bootstrapping` module functions to allow for scaling out if needed (e.g. with dask).

As for block bootstrapping, we want to include an approach that bootstraps whole blocks of a timeseries instead of individual samples from that timeseries. Blocks should be defined based on the correlation time determined from the timeseries, or alternatively specified by the user. There are at least a few ways to do this, so this may require more debate and development. The advantage of block bootstrapping is that the resulting bootstrapped samples maintain correlation within their blocks, which may be desirable in some cases.

### DIFF
--- a/src/alchemlyb/preprocessing/bootstrapping.py
+++ b/src/alchemlyb/preprocessing/bootstrapping.py
@@ -2,22 +2,31 @@
 
 """
 import numpy as np
+import pandas as pd
 
-def bootstrap(df):
+
+def bootstrap(df, random_state=None):
     """Bootstrap sample a DataFrame to produce a new DataFrame.
 
     Values are sampled from `df` on a per-lambda-index basis.
+    Order of samples, or blocks of samples, are not preserved.
 
     Parameters
     ----------
     df : DataFrame
         DataFrame to bootstrap sample from.
+    random_state : int, optional
+        Integer between 0 and 2**32 -1 inclusive; fed to `numpy.random.seed`.
+        Running this function on the same data with a specific random seed will
+        produce the same result each time.
 
     Returns
     -------
     DataFrame
         DataFrame with the same shape as `df`, bootstrapped from the values of `df`.
     """
+    np.random.seed(random_state)
+
     df = df.copy()
     lambda_inds = [i for i in df.index.names if not i == 'time']
     
--- a/src/alchemlyb/preprocessing/bootstrapping.py
+++ b/src/alchemlyb/preprocessing/bootstrapping.py
@@ -1,0 +1,50 @@
+"""Functions for bootstrapping from datasets.
+
+"""
+import numpy as np
+
+def bootstrap(df):
+    """Bootstrap sample a DataFrame to produce a new DataFrame.
+
+    Values are sampled from `df` on a per-lambda-index basis.
+
+    Parameters
+    ----------
+    df : DataFrame
+        DataFrame to bootstrap sample from.
+
+    Returns
+    -------
+    DataFrame
+        DataFrame with the same shape as `df`, bootstrapped from the values of `df`.
+    """
+    df = df.copy()
+    lambda_inds = [i for i in df.index.names if not i == 'time']
+    
+    res = list()
+    for name, group in df.groupby(level=lambda_inds):
+        res.append(_bootstrap_samples(group))
+        
+    return pd.concat(res)
+    
+
+def _bootstrap_samples(df):
+    indices = np.random.choice(len(df), size=len(df))
+    return df.iloc[indices]
+
+
+def block_bootstrap(df, block_size=None, correlation_method=None):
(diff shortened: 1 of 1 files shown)


dotsdl:
Added a [gist](https://gist.github.com/dotsdl/0b0bb7b869acd59dbc4446e0b416542d) on how this works currently. Comments welcome!

orbeckst:
I already like the gist and the dask example. However, shouldn't you decorrelate your data first and then bootstrap the decorrelated data? 

For the "blocks" approach: Do I have to think about a trajectory of blocks where each block is  the slice information to get the original data, something like `slices = [[start0, stop0, step0], [start1, stop1, step1], ..., [startN_1, stopN_1, stepN_1]]` so that `data[slices]` produces N blocks. Then you bootstrap the `indices = arange(len(slices))` of `slices` to reassemble a subset of the original data as `data[slices[bootstrapped_indices]]`?

mrshirts:
> I already like the gist and the dask example. However, shouldn't you decorrelate your data first and then bootstrap the decorrelated data?

For vanilla bootstrap, yes.

For block bootstrap, no -- though to determine the length of the blocks, you do need to know the correlation length.

orbeckst:
If I read the gist correctly then there "vanilla bootstrap" is performed so if the gist is turned into documentation then data decorrelation should probably included so that copy&paste will at least produce conservative estimates.

From the description in the issue above I still don't understand how block bootstraps will be used. A simple example might be helpful.

dotsdl:
@orbeckst thanks! My intent is for the gist to become our actual example for bootstrap usage in the docs, so I am working on adding use of `statistical_inefficiency` in the loop for the **Bootstrapping + estimation** section onward.

However, it's looking like this isn't straightforward with the current implementation limitations of our subsamplers. I think it may be due to a combination of bitrot by way of changes in how `pandas` behaves with multi-index resetting and my rather limiting assumptions when I wrote the subsamplers way back when. I'm going to proceed with getting this right by making the subsampling functions work just fine with full DataFrames like what we get from `alchemtest.gmx.load_expanded_ensemble_case_1`. We need this since it is proving to be too painful `pandas`-fu to work around it. This will be a separate PR.

I believe you have the right idea on how block bootstrapping works. From @mrshirts, there are a few variations on it, however. [Here](http://hrcak.srce.hr/index.php?show=clanak&id_clanak_jezik=197357&lang=en) is a reference I am working from.

dotsdl:
The [gist](https://gist.github.com/dotsdl/0b0bb7b869acd59dbc4446e0b416542d) for has been updated; it requires components of #98, which can be played with on [this branch](https://github.com/alchemistry/alchemlyb/tree/adhoc-tests-subsampling-bootstrapping). This addresses @orbeckst's comments on decorrelation for when this notebook becomes the first example doc for bootstrapping.

orbeckst:
I think pymbar 4 now contains bootstrapping natively (did I understand this correctly, @mrshirts ?). Is this PR still needed when we move to using pymbar 4.x ?

mrshirts:
GREAT question. If you are bootstrapping a single free energy calculation, or a single expectation. pymbar 4 handles it all, and faster than alchemlyb could wrapping around MBAR.

If you were doing something like a heat capacity calculation with multiple temperatures reweighting, or where you do some manipulation of the results of multiple free energy calculations, then you really should bootstrap on the outside of that.  Entropy + enthalpy are done internally.  But if for whatever reason you were calculating k_1 <O_1> - k_2<O_2> , where <O_1> and <O_2> are expectations of different observables from the same simulation and k_1 and k_2 are constants, then you can't bootstrap them separately and add the uncertainties together - you should bootstrap the entire calculation.

If all the functions just are computing a vector f_k from a single single set of files that have been subsampled - then the internal boostrapping should be fine.

orbeckst:
PR #322 addressed #320 so we can now use MBAR-native bootstrap.

I feel that this is sufficient for the moment, especially as TI can get error estimates by other means.

Therefore, I am closing this PR, although of course anyone is welcome to resurrect it and take up work on it again.

Issue #95, opened 2020-01-24
A problem with slicing in subsampling 


wehs7661:
Hi `alchemlyb` developers,
Thank you all for the development of this package. I was using `alchemlyb` to analyze some output files from the simulation performed by GROMACS. When using the function `equilibrium_detection` with the input parameters `df` and `series` being the whole `u_nk` matrix extracted from a `*dhdl.xvg` file (by the function `extract_u_nk`) and the first column of `u_nk` (which is `u_nk[(0.0, 0.0, 0.0, 0.0)]`), I got the following error:
```
KeyError: Index(['time'], dtype='object')
```
Tracing back the error, I found that this error indicated that there was no `time` index in the series `u_nk[(0.0, 0.0, 0.0, 0.0)]` . However, `series.index` in this case did return `time` as one of the names of the indices. I am not sure if this is an issue of data types. I tried turning the data type of `u_nk[(0.0, 0.0, 0.0, 0.0)]` from Series into DataFrame, but same error occurred. Here I reproduced the error in a jupyter notebook, in which I also added some comments and explanations of the problem([alchemlyb_issue.zip](https://github.com/alchemistry/alchemlyb/files/4109820/alchemlyb_issue.zip)).

To my understanding, the input parameter `df` of the subsampling methods should be the matrix/DataFrame extracted from the `xvg` file, while `series` could be any time series/column of `df`. I'm not sure if this understanding is correct, but it might be worthy to add more detailed explanations of these two parameters in the documentation.

Thanks a lot!

dotsdl:
Thank you @wehs7661! I think this may be due to a change in how multi-indexes are handled in `pandas` when resetting indices on a `Series` (see my [comment](https://github.com/alchemistry/alchemlyb/pull/94#issuecomment-577994716)); I have yet to investigate when this change may have happened in `pandas`' release history, but it must be fixed regardless.

This part of the library is in need of some updates to fix this issue, as well as to make the subsamplers easier to use for `u_nk` and `dHdl` dataframes with multiple timeseries concatenated together and having multiple `-lambda` indexes. When the subsamplers were originally written, we had a much smaller set of datasets in [`alchemtest`](https://github.com/alchemistry/alchemtest), and so the assumptions of their implementations are currently too constricted. They currently cannot handle expanded ensemble cases without a lot of pandas-fu beforehand, even if the error above is fixed.

[...]

Issue #91, opened 2019-12-03
Subsampling documentation


Bernadette-Mohr:
Would it be possible to add examples (ideally for multidimensional df) to the documentation of alchemlyb.preprocessing.subsampling?
At least I am not able to guess the usage of statistical_inefficiency() correctly.

dotsdl:
Hi @Bernadette-Mohr, welcome to alchemlyb!

I agree that this is sorely needed. It's been on my radar for some time, but your immediate need pushes the priority up for this. I have added producing a PR addressing #63 to my list for this weekend; this should serve as a good concrete example addressing for this issue.

Are you interested in providing feedback once the PR is up?

Bernadette-Mohr:
Of course, I'll be glad to help!
And yes, generally adding examples for multidimensional dataframes would be awesome.

Bernadette-Mohr:
Hi @dotsdl ,

Thank you for adding the multi-lambda example to the parser documentation! It confirms what I thought the contents were.
But it still doesn't help me with my original question: How would I extract uncorrelated samples from this dataframe with the subsampling module?
No matter what I tried, it either throws an error or returns me the same 2-3 rows for every lambda step, so probabely just acting like slicing.

Bernadette-Mohr:
Sorry to complain again, but a hint at how to extract the individual contributions of VdW and Coulomb interactions would be helpful too.
I guess I would have to split the multiindex dataframe by coul-lambda=1.0 AND vdw-lambda != 0.0, but that's just a guess.

Issue #79, opened 2019-05-02
Subsampling


msoroush:
In order to use subsampling functions, `time` must be used as index name of data frame. In GOMC parser, I am using `step` as index name, which I receive error when I want to use subsampling functions, such as slicing. 
```
KeyError: Index(['time'], dtype='object')
```
One possible solution is to change index name to `time` in GOMC parser. However, generalizing the subsampling module to work with any index name might be useful.

Is it possible to have a function to perform slicing on Concatenate data frame?

orbeckst:
The dataframes are standardized as described in [standard forms of raw data](https://alchemlyb.readthedocs.io/en/latest/parsing.html#standard-forms-of-raw-data) and there "time" has to be included. The standardization is crucial to make all parts of alchemlyb work seamlessly together and changing it might not be easy.

I see two options:

1. You could rename `step` to `time` in PR #78. You should add tests that show that the parsed data can be processed with subsampling and estimators.
2. You could propose a change to all the subsampling and estimator modules to be more flexible with respect to the "time" column. For that, submit a PR that shows that this change is not breaking everything else and we review it then. I won't promise that it will get merged – it looks as if could be a big change with a narrow range of applicability. But that's a discussion we would have on the PR when we see actual code and tests.

(You could argue that we should have included a column `frame` or `step` in addition to `time` in the raw data frame format. However, so far the lack of this column has not prevented the library from doing what it's supposed to. One would need examples where this information would be necessary.)

orbeckst:
Just to add: it would be great to make alchemlyb also work with MC data.

Issue #34, opened 2017-11-01
use logging


orbeckst:
For debug and info output we do not want to use `print` but use the standard [logging](https://docs.python.org/3/howto/logging.html) facility.

orbeckst:
See https://github.com/alchemistry/alchemlyb/pull/32#discussion_r148339627 for more discussion.

orbeckst:
With PR #303 we now have docs for logging.

PR #98, opened 2020-01-29
WIP: Refactor of subsampling module

dotsdl:
Addresses #95, #79, #91 

Required for merge:
- [x] reference documentation for the functions included in the `alchemlyb.preprocessing.subsampling` module
- [ ] example notebook showing usage, as well as rationale for usage, of each function
- [x] tests across more datasets in `alchemtest`
- [ ] changelog entry

This is an API breaking change.

All three of our subsampling functions should now be able to consume any DataFrame in our two standard `u_nk` and `dHdl` forms and produce meaningful results. Each of these functions now performs a `groupby` with all non-time indexes, and separately performs its operation on each group. The resulting subsampled groups are concatenated and returned.

Some notes on the implementation:
1. The `series` kwarg is now the `column` kwarg for  `statistical_inefficiency`, `equilibrium_detection`. This can take either a column name in the DataFrame or a Series object (the index of the Series must match *exactly* that of the DataFrame). Taking a Series object allows for code that uses the existing usage of e.g. `statistical_inefficiency(u_nk, u_nk[<column_name>])` to still work as before.
2. `statistical_inefficiency` and `equilibrium_detection` now have a `how` kwarg. I expect this to be the main usage we suggest for users of these functions. I took a look at how `alchemical_analysis` [currently handles `dHdl` and `u_nk` cases](https://github.com/MobleyLab/alchemical-analysis/blob/master/alchemical_analysis/alchemical_analysis.py#L159), and to replicate that as closely as possible, we recommend using `how='right'` for `u_nk` data, and `how='sum'` for `dHdl` data. Please see the docstrings for `statistical_inefficiency`, `equilibrium_detection` for details on these treatments.
3. Either `column` or `how` must be used for `statistical_inefficiency`, `equilibrium_detection`. The function can no longer default to simple slicing if only `data` is given as an input. This is to avoid (potentially new) users from applying the functions without specifying these inputs with no obvious failure, but with the function doing nothing to decorrelate their data (essentially a silent failure).
4. I've added the `force` kwarg to `statistical_inefficiency`, `equilibrium_detection`. The only exception these functions will skip with `force=True` is one due to duplicate time values in the groupings, as this tends to indicate two or more timeseries have been dumped together, which is unlikely what one wants to do with these subsampling functions that assume correlation to work as intended.
5. Setting `return_calculated=True` for `statistical_inefficiency`, `equilibrium_detection` will return calculated values used for subsampling each lambda grouping:
    - `statistical_inefficiency`: `statinef`
    - `equilibrium_detection`: `t`, `statinef`, `Neff_max`

Some other nice things:
1. I will be adding the minimal assumptions the standard forms make to their doc page as part of this change's PR. One of the assumptions I want to propagate is that the outermost index is *always* time, or at least something implying time (frame, sample, whatever). It needn't be called 'time', however, for all the tooling to work.
2. These subsamplers now explicitly call out that they *can't be expected* to preserve any custom ordering of rows. They will always be doing sorting of some kind, so the time index values are meaningful.

### DIFF
--- a/docs/preprocessing/alchemlyb.preprocessing.subsampling.rst
+++ b/docs/preprocessing/alchemlyb.preprocessing.subsampling.rst
@@ -1,3 +1,5 @@
+.. currentmodule:: alchemlyb.preprocessing.subsampling
+
 .. _subsampling:
 
 Subsampling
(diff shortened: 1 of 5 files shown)


dotsdl:
This [gist](https://gist.github.com/dotsdl/a526a5827781f6e9c05f229e81487b06) demonstrates the behavior of the refactored functions.

dotsdl:
The changes I've made to the codebase are ready for review! I've updated the PR description with the current state of the design, as it took a few iterations for me to arrive at. I would like some feedback on this design as soon as possible so changes can be made to accommodate needs. This PR is also a bit of a blocker for #94 and #99, since subsampling is a key component of the `alchemlyb` workflow.

I am now working on an example notebook that will make its way into our examples docs; this will showcase how to use these functions, as well as when each makes sense to use. I don't consider this notebook a requirement for review of the implementation, but consider it important to make sure the feel of this approach is right. I think I can turn it around and add to this PR by the end of this week.

dotsdl:
@orbeckst and @mrshirts I've requested your eyes to give this implementation a pass. In particular, I would like a check on the `how` parameter options for `statistical_inefficiency` and `equilibrium_detection`, and whether they make sense or appear correct.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -1,26 +1,79 @@
...
+
+    return group[group.columns[pos]]
+
+
+def _how_random(name, group, tried=None):
+
+    candidates = set(group.columns) - (set(tried) | {name})
+
+    if not candidates:
+        raise CorrelationError("No column in the dataset could be used"
```
orbeckst: This could be `RuntimeError`.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -34,63 +87,198 @@ def slicing(df, lower=None, upper=None, step=None, force=False):
...
+    Calculation of statistical inefficiency and subsequent subsampling will be
+    performed separately on groups of rows corresponding to each set of lambda
+    values present in the DataFrame's index. Each group will be sorted on the
+    outermost (time) index prior to any calculation.
+
+    The `how` parameter determines the observable used for calculating the
+    correlations within each group of samples. The options are as follows:
+
+        'right'
+            The recommended setting for 'u_nk' datasets; the column immediately
```
orbeckst: Who's doing the recommending and on what grounds?

If my student used alchemlyb and said "I used the recommended settings" then I will ask "recommendation based on what?" and then the student won't know what to say.

I'd like a more scholarly approach to recommendations. Or don't recommend anything.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -34,63 +87,198 @@ def slicing(df, lower=None, upper=None, step=None, force=False):
...
+            index value is used. If there is no column to the right, then the
+            column to the left is used.  If there is no column corresponding to
+            the group's lambda index value, then 'random' is used (see below).
+        'left'
+            The opposite of the 'right' approach; the column immediately to the
+            left of the column corresponding to the group's lambda index value
+            is used. If there is no column to the left, then the column to the
+            right is used.  If there is no column corresponding to the group's
+            lambda index value, then 'random' is used for that group (see below).
+        'random'
```
orbeckst: Why this option?? This seems counterproductive to reproducible research. Does the user even know which column worked in the end? I'd much rather fail.

orbeckst: I really do not like this option in a library that tries to be clean and basic. Let the users write their workflows and heuristics. 👎

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -34,63 +87,198 @@ def slicing(df, lower=None, upper=None, step=None, force=False):
...
+        ``True`` use ``ceil(statistical_inefficiency)`` to slice the data in
+        uniform intervals (the default). ``False`` will sample at non-uniform
+        intervals to closely match the (fractional) statistical_inefficieny, as
+        implemented in :func:`pymbar.timeseries.subsampleCorrelatedData`.
+    return_calculated : bool
+        ``True`` return a tuple, with the second item a dict giving, e.g. `statinef`
+        for each group.
+    force : bool
+        Ignore checks that DataFrame is in proper form for expected behavior.
+    random_state : int
```
orbeckst: 👎

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -34,63 +87,198 @@ def slicing(df, lower=None, upper=None, step=None, force=False):
...
 
-    Warning
-    -------
-    The `series` and the data to be sliced, `df`, need to have the same number
-    of elements because the statistical inefficiency is calculated based on
-    the index of the series (and not an associated time). At the moment there is
-    no automatic conversion from a time to an index.
+    Raises
+    ------
+    CorrelationError
```
orbeckst: Not sure if custom error is justified.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -102,101 +290,208 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None,
     decreases a false sense of accuracy and is deemed the more careful and
     conservative approach.
 
+    References
+    ----------
+    [1] J. D. Chodera, W. C. Swope, J. W. Pitera, C. Seok, and K. A. Dill. Use of the weighted
```
orbeckst: I would use reST citations and have one central citations page. At the end of the day, I think users are more likely to read the online docs for this kind of information than when they read the raw doc strings.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -102,101 +290,208 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None,
...
+       The ``series`` keyword was replaced with the ``column`` keyword.
+       The function no longer takes an arbitrary series as input for
+       calculating statistical inefficiency.
+
     .. versionchanged:: 0.2.0
        The ``conservative`` keyword was added and the method is now using
-       ``pymbar.timeseries.statisticalInefficiency()``; previously, the statistical
-       inefficiency was _rounded_ (instead of ``ceil()``) and thus one could
-       end up with correlated data.
+       ``pymbar.timeseries.subsampleCorrelatedData()``; previously, the
```
orbeckst: Did pymbar change the function name or was this a mistake?

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -102,101 +290,208 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None,
...
-    :func:`slicing`.
+        indices = timeseries.subsampleCorrelatedData(group_cs, g=statinef,
+                                      conservative=conservative)
+
+        calculated = dict()
+        calculated['statinef'] = statinef
+
+        return indices, calculated
+
+    return _decorrelator(
```
orbeckst: I like the more modular approach here.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -102,101 +290,208 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None,
...
+            index value is used. If there is no column to the right, then the
+            column to the left is used.  If there is no column corresponding to
+            the group's lambda index value, then 'random' is used (see below).
+        'left'
+            The opposite of the 'right' approach; the column immediately to the
+            left of the column corresponding to the group's lambda index value
+            is used. If there is no column to the left, then the column to the
+            right is used.  If there is no column corresponding to the group's
+            lambda index value, then 'random' is used for that group (see below).
+        'random'
```
orbeckst: Not a fan... 👎

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -10,117 +10,284 @@
...
+                                 (True, gmx_benzene_u_nk(), 2001, 'right'),  # 'fep': g = 1.0560203916559594
+                                 (False, gmx_benzene_dHdl(), 3789, 'sum'),
+                                 (False, gmx_benzene_u_nk(), 3788, 'right'),
+                             ])
+    def test_conservative(self, data, size, conservative, how):
+        sliced = self.subsampler(data, how=how, conservative=conservative)
+        # results can vary slightly with different machines
+        # so possibly do
+        # delta = 10
+        # assert size - delta < len(sliced) < size + delta
```
orbeckst: You could also turn it into a floating point comparison, e.g., `size/len(sliced)` as the fraction and then just use `assert_almost_equal(...., decimals=2)` or whatever. Either way, if you already know that there can be variation then put it in the test and comment it.

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -10,117 +10,284 @@
...
-                                 (False, gmx_benzene_u_nk(), 3571),
+                                 (True, gmx_benzene_dHdl(), 2001, 'sum'),  # 0.00:  g = 1.0559445620585415
+                                 (True, gmx_benzene_u_nk(), 2001, 'right'),  # 'fep': g = 1.0560203916559594
+                                 (False, gmx_benzene_dHdl(), 2844, 'sum'),
+                                 (False, gmx_benzene_u_nk(), 2845, 'right'),
                              ])
-    def test_conservative(self, data, size, conservative):
-        sliced = self.slicer(data, series=data.iloc[:, 0], conservative=conservative)
+    def test_conservative(self, data, size, conservative, how):
+        sliced = self.subsampler(data, how=how, conservative=conservative)
```
orbeckst: less stringent test?

orbeckst (commented):
Hi @dotsdl , I had a quick look and overall this looks like great work. Some comments inline.

docs/preprocessing/alchemlyb.preprocessing.subsampling.rst
```diff
@@ -1,11 +1,268 @@
...
+
+Statistical Inefficiency
+------------------------
+The :func:`~alchemlyb.preprocessing.subsampling.statistical_inefficiency` function subsamples each timeseries in the dataset by its calculated *statistical inefficiency*, :math:`g`, defined as:
+
+.. math::
+
+    g \equiv (1 + 2\tau) > 1
+
+where :math:`\tau` is the autocorrelation time of the timeseries.
```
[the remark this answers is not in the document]
dotsdl: Ah, good call. Tau in this usage is measured in frames, or ordered rows, of a timeseries. I will change the wording to reflect this.

docs/preprocessing/alchemlyb.preprocessing.subsampling.rst
```diff
@@ -1,11 +1,268 @@
...
+    39940.0 1.0        -1.480193 -1.034104 -0.588015  -0.141925   0.304164
+    39960.0 1.0         3.913594  3.011197  2.108801   1.206405   0.304009
+    39980.0 1.0         1.495407  1.199280  0.903152   0.607024   0.310897
+    40000.0 1.0         0.715197  0.611187  0.507178   0.403169   0.299160
+    
+    [12005 rows x 5 columns]
+
+The ``how`` parameter indicates the choice of observable for performing the calculation of :math:`g`.
+
+* For :math:`u_{nk}` datasets, the choice of ``'right'`` is recommended: the column immediately to the right of the column corresponding to the group's lambda index value is used as the observable.
```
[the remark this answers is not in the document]
dotsdl: I'd be happy to, but they come from my attempt to emulate what is already done in [alchemical-analysis.py](https://github.com/MobleyLab/alchemical-analysis/blob/master/alchemical_analysis/alchemical_analysis.py#L159), and I don't know if a reference exists for the choices made there. @mrshirts may be able to provide some direction on this.

I do believe I understand the reason for the choices, however: in principle, any column in a timeseries would be fine to base decorrelation on, so long as the column is not uniformly zero or insensitive to time/frame for the given lambda window being sampled. 

1. The approach of choosing the column to the right of the lambda window being sampled for `u_nk` avoids cases where the `u_nk` of calculated lambdas are relative to the sampled one, giving all zeros for the column corresponding to the lambda actually sampled. An example of this observed:

```python
from alchemtest.gmx import load_water_particle_without_energy
import pandas as pd

ds = load_water_particle_without_energy()

print(pd.concat(extract_u_nk(i, T=300) for i in ds.data['AllStates']).sort_index(level=[1, 2, 0]))
```
This gives:
```
                               (0.0, 0.0)  ...    (1.0, 1.0)
time   coul-lambda vdw-lambda              ...              
0.0    0.0         0.0           0.000000  ...    148.347412
18.6   0.0         0.0           0.000000  ...   1983.809792
37.2   0.0         0.0           0.000000  ...    205.930556
55.8   0.0         0.0           0.000000  ...   4512.816690
74.4   0.0         0.0           0.000000  ...  25939.040202
...                                   ...  ...           ...
9913.8 1.0         1.0          45.353152  ...      0.000000
9932.4 1.0         1.0          41.207448  ...      0.000000
9951.0 1.0         1.0          44.375182  ...      0.000000
9969.6 1.0         1.0          45.505032  ...      0.000000
9988.2 1.0         1.0          37.779798  ...      0.000000

[20444 rows x 38 columns]


```
Note that if we chose any old column, for one of the groupings of the `('coul-lambda', 'vdw-lambda`)` indices, we would end up trying to decorrelate on all zeros. the `how="right"` choice sidesteps this (as does `how="left"`).

2. The approach of summing all `dHdl` columns is not as clear to me, and I wouldn't be surprised if I've misinterpreted the implementation in `alchemical-analysis.py`.

docs/preprocessing/alchemlyb.preprocessing.subsampling.rst
```diff
@@ -1,11 +1,268 @@
...
+* For :math:`\frac{dH}{d\lambda}` datasets, the choice of ``'sum'`` is recommended: the columns are simply summed, and the resulting :py:class:`pandas.Series` is used as the observable.
+
+See the API documentation below on the possible values for ``how``, as well as more detailed explanations for each choice.
```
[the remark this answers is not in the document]
dotsdl: Agreed, I would like some assistance with this. See my notes above.

docs/preprocessing/alchemlyb.preprocessing.subsampling.rst
```diff
@@ -1,11 +1,268 @@
...
+
+See :py:func:`pymbar.timeseries.statisticalInefficiency` for more details; this is used internally by this subsampler.
+Please reference the following if you use this function in your research:
+
+    [1] J. D. Chodera, W. C. Swope, J. W. Pitera, C. Seok, and K. A. Dill. Use of the weighted histogram analysis method for the analysis of simulated and parallel tempering simulations. JCTC 3(1):26-41, 2007.
+
+
+Equilibrium Detection
+---------------------
+The :func:`~alchemlyb.preprocessing.subsampling.equilibrium_detection` function subsamples each timeseries in the dataset using the equilibration detection approach developed by John Chodera (see reference below).
```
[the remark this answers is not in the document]
dotsdl: Understood, I will change this accordingly.

dotsdl:
Thank you for your review @orbeckst! Your comments are incredibly helpful in steering this refactor to a satisfying state; I will be addressing the points you have raised over the next several days.

src/alchemlyb/preprocessing/subsampling.py
dotsdl: This was a mistake in the docstring that I fixed here; the same mistake was not made in the equivalent line for `equilibrium_detection`.

src/alchemlyb/preprocessing/subsampling.py
dotsdl: Thank you!

src/alchemlyb/preprocessing/subsampling.py
dotsdl: I agree that we would like to avoid randomness where possible. We could remove the `how='random'` parameter choice here, but we may still need to internally address the (admittedly rare?) corner case of a `u_nk` column not existing for a lambda state actually sampled. For the "right"/"left" approaches, if a corresponding column doesn't exist for the lambda state in question, the concept of right and left becomes more guesswork.

Some alternative approaches I can think of:
1. Find the "closest" column in terms of euclidean distance in lambdas, and then go to the right or left of that, but this logic seems to be pretty tortured to me, and I'm not sure if it too can fail in weird ways.
2. Choose the leftmost or rightmost column in the whole DataFrame, and march through columns in order if we hit failures to decorrelate.
3. Try to use the last column we subsampled on (previous iteration of [this loop](https://github.com/alchemistry/alchemlyb/pull/98/files/dc1b3ff106eb1d66f1b13555c6d1abe3d65c0e8d#diff-29b9db00345fef71312e85c1392b6fbeR157)).
4. Fail loudly, and declare this corner case not one we wish to address. The user could get around this by throwing out rows corresponding to lambda values there are no corresponding columns for.

dotsdl: One way we can address some of the harder-to-pin-down choices made in alchemlyb is to use a logger (#34). I recognize that this doesn't excuse *strange* choices on the library's part, but if we have a logger in place then at least what the library did in a particular instance is auditable later.

orbeckst: Logging would be a good step towards more transparency.

Maybe put in a commented-out logger call with a hint towards #34?

dotsdl: Will do! I consider this an important component of the library.

orbeckst:
@dotsdl what do we need here to get this in? I’m asking because if we want to do breaking changes then I’d rather do it sooner than later. 

Or we do 0.4 #113 and then do this one for a 0.5 with an eye towards 1.0. Overall it looks to me that alchemlyb has sufficiently stabilized to just declare the API stable soon.

xiki-tempula:
I don't think there are significant features that are not implemented in the 2.0.0. But I do feel that the documentation could benefit.

Issue #73, opened 2019-03-21
Port in plotting/analysis functionality from alchemical-analysis and yank


davidlmobley:
In recent discussions with OpenEye about free energy calculations on their Orion cloud computing platform, it became clear they will benefit from an open analysis library (`alchemlyb` in particular) because users are running calculations with Yank, GROMACS, and probably soon AMBER on their platform. Thus they are likely prepared to invest some developer time in ensuring `alchemlyb` has the functionality needed. I was recently discussing this with Christopher Bayly ( @cbayly13 ) and Gaetano Calabro (  @nividic ) there and I said I'd summarize what I thought needed to be done on the issue tracker so that @nividic could begin work in the coming weeks if everyone here is on board with it.

I think the main things that I think we would want added to analysis in the short term is:
1) Overlap matrix analysis from `alchemical-analysis`, e.g. as in DOI 10.1007/s10822-015-9840-9
2) Mixing diagrams as seen in Yank's simulation health reports
3) Graphs of equilibration/number of effective samples as in Yank's simulation health reports
4) Numerical analysis of statistical inefficiency/number of effective samples in input datasets 
5) Any other key features of Yank simulation health reports I'm missing? (I don't have a report in front of me at the moment.)

Possibly also there is a need for ensuring at least a threshold minimal number of samples are retained for analysis after decorrelation. 

For Orion purposes I would also suggest running multiple analysis methods (BAR, MBAR, TI) whenever the data allows for it (often) and cross-checking results for consistency; inconsistencies are usually a warning of problems.

An additional list of possible changes is on the [alchemical analysis features](https://github.com/alchemistry/alchemlyb/wiki/alchemical-analysis-features) list where we were brainstorming about this. 

Tagging also @harlor for thoughts as well as @hannahbaumann and @andrrizzi . And of course @dotsdl and @orbeckst .

@nividic the way to get started would be to basically pull the relevant code from `alchemical-analysis` and `yank` (allowable by the licenses) and adapt to this.

mrshirts:
Cool, I'll try to keep posted on this.  Automated bootstrap error (better that the internal error estimator in most cases).

harlor:
I already started point 4. in https://github.com/harlor/alchemlyb/tree/autocorrelation_analysis maybe I'll find time on the weekend to continue.

luancarvalhomartins:
Dear all.

[...]

PR #110, opened 2020-10-01
dhdl plot for the TI

xiki-tempula:
Partially addresses #73

I have ported the dhdl plot for TI from the alchemical analysis.

Based on user experience, I have changed a few things.

- The user could now set custom label to the figure
- The x axis is elevated a bit to avoid bumping to the dhdl
- The user could now custom the color of the plot


xiki-tempula:
@orbeckst I'm sorry that I felt that I cannot bump the test coverage to any higher.
I have ported the code from the alchemical analysis and modified it such that it works in py38.

However, there are two regions that I'm not quite sure of what it does.

```
    def getInd(r=ri, z=[0]):
        primo = r[0]
        min_dl = ndx * 0.02 * 2 ** (primo > 10)
        if dl_mat[primo].max() < min_dl:
            return z
        for i in r:
            for j in range(len(xs)):
                if dl_mat[i, j] > min_dl:
                    z.append(j)
                    return getInd(ri[j:], z)
```
This function exists to make sure that there won't be too many ticks in the x axis.
I have tested this function with
```
    ti_coul.dhdl = pd.DataFrame.from_dict(
        {'fep': range(100)},
        orient='index',
        columns=np.arange(100)/100).T
    assert isinstance(plot_ti_dhdl(ti_coul),
               matplotlib.axes.Axes)
```
but cannot get it fully covered.

The other part is
```
    if ndx > 1:
        lenticks = len(ax.get_ymajorticklabels()) - 1
        if min_y < 0: lenticks -= 1
        if lenticks < 5:
            from matplotlib.ticker import AutoMinorLocator as AML
            ax.yaxis.set_minor_locator(AML())
```
Which my understanding is to deal with the case when there are too tittle ticks in the x axis.
I have tested it with
```
    ti_coul.dhdl = pd.DataFrame.from_dict(
        {'fep': range(3)},
        orient='index',
        columns=np.arange(3)/3).T
    assert isinstance(plot_ti_dhdl(ti_coul),
               matplotlib.axes.Axes)
```
but cannot get it fully covered.

orbeckst (changes requested):
Nice work and kudos for pushing up coverage as high as you did – it's hard with plots!

Given that you identified the code pieces that are hard to test, mark them explicitly as not included in coverage. Something like
```python
# pragma: no cover
```
You might have to find out how exactly to do this (we do it in MDAnalysis) but with this you explicitly acknowledge that certain code is untested and then it's removed from the coverage calculation.

xiki-tempula:
@orbeckst Thank you for the advice. I have commented out the relevant lines with `# pragma: no cover`.

src/alchemlyb/visualisation/ti_dhdl.py
```diff
@@ -0,0 +1,167 @@
...
+        spine.set_zorder(12.2)
+
+    # Make level names
+    if labels is None:
+        lv_names2 = []
+        for dhdl in dhdl_data:
+            # Assume that the dhdl has only one columns
+            lv_names2.append(dhdl.dhdl.columns.values[0].capitalize())
+    else:
+        assert len(labels) == len(dhdl_data), \
```
orbeckst, line 72: Change this `assert` to a `raise ValueError`.

We should not be using assertions to capture user errors.

src/alchemlyb/visualisation/ti_dhdl.py
```diff
@@ -0,0 +1,167 @@
...
+            'Length of labels ({}) should be the same as the number of data ({})'.format(len(labels), len(dhdl_data))
+        lv_names2 = labels
+
+    if colors is None:
+        colors = ['r', 'g', '#7F38EC', '#9F000F', 'b', 'y']
+    else:
+        assert len(colors) >= len(dhdl_data), \
```
orbeckst, line 79: `raise ValueError` instead

If this is not tested, use `#pragma: no cover`

src/alchemlyb/visualisation/ti_dhdl.py
orbeckst: If this is not tested, use `#pragma: no cover`

orbeckst (changes requested):
Thanks for adjusting coverage. I just saw two things to change: raise exceptions instead of using assertions for input checking – this was probably already in alchemical-analysis but should certainly not be in alchemlyb.

Otherwise, this looks excellent.

orbeckst (approved):

orbeckst:
Thank you @xiki-tempula !! 🚀

PR #119, opened 2021-03-27
Add sort and remove duplication to statistical_inefficiency

### LINKED ISSUE #118: Adding automatic sorting and duplicate removal
Currently, the preprocessing.subsampling.statistical_inefficiency will reject the data set if it is not sorted and contains duplications.
I'm aware that @dotsdl is refactoring the subsampling module #98.

I think having the automatic sorting and duplication removal functionality will make the ABFE workflow more tolerable to corrupted datasets.

I'm thinking of adding the sorting and duplication removal to the preprocessing.subsampling.statistical_inefficiency.

```
def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None,
                             conservative=True, drop_duplicates=True, sort=True):
    """Subsample a DataFrame based on the calculated statistical inefficiency
    of a timeseries.

    If `series` is ``None``, then this function will behave the same as
    :func:`slicing`.

    Parameters
    ----------
    df : DataFrame
        DataFrame to subsample according statistical inefficiency of `series`.
    series : Series
        Series to use for calculating statistical inefficiency. If ``None``,
        no statistical inefficiency-based subsampling will be performed.
    lower : float
        Lower bound to pre-slice `series` data from.
    upper : float
        Upper bound to pre-slice `series` to (inclusive).
    step : int
        Step between `series` items to pre-slice by.
    conservative : bool
        ``True`` use ``ceil(statistical_inefficiency)`` to slice the data in uniform
        intervals (the default). ``False`` will sample at non-uniform intervals to
        closely match the (fractional) statistical_inefficieny, as implemented
        in :func:`pymbar.timeseries.subsampleCorrelatedData`.
    drop_duplicates : bool
        Drop the duplicated lines based on time.
    sort : bool
        Sort the Dataframe based on the time column.
    """
```
IOr it is better to make it functionality of the ABFE workflow?

xiki-tempula:
Fix #118

### DIFF
--- a/src/alchemlyb/tests/test_preprocessing.py
+++ b/src/alchemlyb/tests/test_preprocessing.py
@@ -78,6 +78,31 @@ class TestSlicing:
                                              gmx_ABFE.sum(axis=1))
         assert len(subsample) == 501
 
+    def test_sort(self, gmx_ABFE):
+        unsorted = pd.concat([gmx_ABFE[-500:], gmx_ABFE[:500]])
+        # No sort
+        with pytest.raises(KeyError):
+            statistical_inefficiency(unsorted,
+                                     unsorted.sum(axis=1),
+                                     sort=False)
+        # sort
+        subsample = statistical_inefficiency(unsorted,
+                                             unsorted.sum(axis=1),
+                                             sort=True)
+        assert subsample.reset_index(0)['time'].is_monotonic_increasing
+
+    def test_duplication(self, gmx_ABFE):
+        duplicated = pd.concat([gmx_ABFE, gmx_ABFE])
+        # No sort
+        with pytest.raises(KeyError):
+            statistical_inefficiency(duplicated,
+                                     duplicated.sum(axis=1),
+                                     drop_duplicates=False)
+        # sort
+        subsample = statistical_inefficiency(duplicated,
+                                             duplicated.sum(axis=1),
+                                             drop_duplicates=True)
+        assert len(subsample) < 1000
 
 class CorrelatedPreprocessors:
 
@@ -135,3 +160,4 @@ class TestEquilibriumDetection(TestSlicing, CorrelatedPreprocessors):
 
     def slicer(self, *args, **kwargs):
         return equilibrium_detection(*args, **kwargs)
+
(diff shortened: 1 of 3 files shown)


src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -78,6 +78,37 @@ def test_multiindex_duplicated(self, gmx_ABFE):
                                              gmx_ABFE.sum(axis=1))
         assert len(subsample) == 501
 
+    def test_sort(self, gmx_ABFE):
```
orbeckst, line 81: split in two tests, one for unsorted and KeyError, the other for sorted. More granular tests are better and failing the first stops testing the second in the current form

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -78,6 +78,37 @@ def test_multiindex_duplicated(self, gmx_ABFE):
...
+            statistical_inefficiency(unsorted,
+                                     unsorted.sum(axis=1),
+                                     sort=False)
+        # sort
+        subsample = statistical_inefficiency(unsorted,
+                                             unsorted.sum(axis=1),
+                                             sort=True)
+        assert subsample.reset_index(0)['time'].is_monotonic_increasing
+
+    def test_duplication(self, gmx_ABFE):
```
orbeckst, line 94: split in three tests

orbeckst (changes requested):
Overall this looks sensible to me. Maybe @dotsdl wants to have a quick look.

I mainly have comments on the tests.

Also add an entry to CHANGELOG, please.

xiki-tempula:
@orbeckst Thank you for the review. I have updated the PR.

orbeckst (approved):

orbeckst:
@xiki-tempula , this looks good to me. But do you know why the coverage is so low? I see that two `if` statements are not fully checked (essentially, only the `if True` code path is tested while the effective `else: pass` is not). Under which circumstances would these other code paths be triggered?

xiki-tempula:
@orbeckst Thank you for the review. The two if checks if the `series` is given and will only modify the series if the series is given.
If the series is not given, the data will be sorted and duplications removed but not decorrelated.

orbeckst:
Then these should be simple tests: just as before, just don't give the series. It seems a bit over the top but could you please add tests for this case, too? Then this PR will be done.

xiki-tempula:
@orbeckst Thanks for the suggestions, I have bumped the test.

PR #120, opened 2021-03-29
separate_dhdl

xiki-tempula:
It might be useful to separate the dataframe given by the TI.dhdl when multiple stages transformations are used into a list of series where each series is one stage transformation.


src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,45 @@ def fit(self, dHdl):
...
+        Returns
+        ----------
+        dHdl_list : list
+            A list of Series such that dHdl_list[k][n] is the potential
+            energy gradient with respect to lambda for each configuration n and
+            lambda k.
+        """
+        if len(self.dhdl.index.names) == 1:
+            # If only one column is present convert to series
+            assert len(self.dhdl.columns) == 1
```
orbeckst, line 123: Don't use `assert` for error checking. This can be optimized away with `python -O`.

If you want to check for error, use try/except or if and raise the appropriate exception, e.g., `ValueError`.

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,45 @@ def fit(self, dHdl):
...
+            name = self.dhdl.columns[0]
+            return [self.dhdl[name], ]
+        else:
```
orbeckst, line 126: Remove the `else`. The `return` in the previous block terminates control flow. Dedent the following block to make it more readable. It's a bit ugly if the main code in a function is in an `else` block.

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,45 @@ def fit(self, dHdl):
...
+            l_types = self.dhdl.index.names
+            # obtain bool of changed lambdas between each state
+            lambdas = self.dhdl.reset_index()[l_types]
+            diff = lambdas.diff().to_numpy(dtype='bool')
+            # diff will give the first row as NaN so need to fix that
+            diff[0, :] = diff[1, :]
+            # Make sure that the start point is set to true as well
+            diff[:-1, :] = diff[:-1, :] | diff[1:, :]
+            for i in range(len(l_types)):
+                if any(diff[:,i]) == False: # pragma: no cover
```
orbeckst, line 138: Perhaps rewrite so that you don't need `pass` and just use a `not any(....)`?

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,45 @@ def fit(self, dHdl):
         self.states_ = means.index.values.tolist()
 
         return self
+
+    def separate_dhdl(self):
+        """
+        For transitions with multiple lambda, this function will separate the
+        dhdl with multiple columns into a list of Dataframe with a single column
+        (single lambda).
```
orbeckst, line 113: add empty line before Return section.

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,45 @@ def fit(self, dHdl):
         self.states_ = means.index.values.tolist()
 
         return self
+
+    def separate_dhdl(self):
+        """
+        For transitions with multiple lambda, this function will separate the
+        dhdl with multiple columns into a list of Dataframe with a single column
```
orbeckst, line 112: Explain more clearly what the list contains. Is this a list of multiple DataFrame instances where each DataFrame contains ... what?

Explain it to someone who does not have the data structure in front of themselves.

Mention why this might be useful.

src/alchemlyb/tests/test_ti_estimators.py
```diff
@@ -136,3 +136,13 @@ class TestTI(TIestimatorMixin):
     def X_delta_f(self, request):
         get_dHdl, E, dE = request.param
         return get_dHdl(), E, dE
+
+def test_TI_separate_dhdl_multiple_column():
+    dHdl = gomc_benzene_dHdl()
+    estimator = TI().fit(dHdl)
+    assert all([isinstance(dhdl, pd.Series) for dhdl in estimator.separate_dhdl()])
```
orbeckst, line 143: These tests should also check correctness, i.e., values or aggregate values.

src/alchemlyb/tests/test_ti_estimators.py
```diff
@@ -136,3 +136,13 @@ class TestTI(TIestimatorMixin):
...
+
+def test_TI_separate_dhdl_single_column():
+    dHdl = gmx_benzene_coul_dHdl()
+    estimator = TI().fit(dHdl)
+    assert all([isinstance(dhdl, pd.Series) for dhdl in estimator.separate_dhdl()])
```
orbeckst, line 148: These tests should also check correctness, i.e., values or aggregate values.

orbeckst (changes requested):
Please see comments. Apologies for brevity.

xiki-tempula:
@orbeckst Thank you for the review. I have updated the PR. I noticed that the convergence analysis will raise a problem if the else statement is not checked.

```
if true:
    pass
else:
    pass
```

if the else is not present, it will raise it as not covered by the test.

src/alchemlyb/estimators/ti_.py
orbeckst: I read your comment regarding coverage problems. Nevertheless, can you please dedent the whole `else` block and then we see what happens.

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,50 @@ def fit(self, dHdl):
...
+        if len(self.dhdl.index.names) == 1:
+            # If only one column is present convert to series
+            if len(self.dhdl.columns) == 1:
+                name = self.dhdl.columns[0]
+                return [self.dhdl[name], ]
+            else: # pragma: no cover
+                raise ValueError('The index column and value column should have'
+                                 ' the same length. Index: {}; Value: {}'
+                                 ''.format(len(self.dhdl.index.names),
+                                           len(self.dhdl.columns)))
```
orbeckst, line 133: Under which circumstances do you expect this exception to be triggered?

I'd be happier if it gets tested. alchemlyb is trying really hard to have everything tested and experience shows that untested code leads to problems — typically for other people than the author so I'd rather have the author do the work up-front than someone else having to fix it later.

Perhaps you can write a test that provides a very simple dataframe that triggers the exception or you can modify one of the standard test cases in such a way that it fails.

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,50 @@ def fit(self, dHdl):
...
+        For transitions with multiple lambda, the `dhdl` attribute would return
+        a DataFrame which gives the dHdl for all the lambda states, regardless
+        of whether it is perturbed or not. This function creates a list of
+        Series for each lambda, where each Series describes the potential energy
+        gradient for the lambdas state that is perturbed.
+
+        Returns
+        ----------
+        dHdl_list : list
+            A list of Series such that dHdl_list[k] is the potential
```
orbeckst, line 120: Markup code
```suggestion
            A list of :class:`pandas.Series` such that ``dHdl_list[k]`` is the potential
```

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,50 @@ def fit(self, dHdl):
         self.states_ = means.index.values.tolist()
 
         return self
+
+    def separate_dhdl(self):
+        """
+        For transitions with multiple lambda, the `dhdl` attribute would return
```
orbeckst, line 111: Use rich reST markup:
```suggestion
        For transitions with multiple lambda, the attr:`dhdl` attribute would return
```

src/alchemlyb/estimators/ti_.py
```diff
@@ -105,3 +105,50 @@ def fit(self, dHdl):
...
+        a DataFrame which gives the dHdl for all the lambda states, regardless
+        of whether it is perturbed or not. This function creates a list of
+        Series for each lambda, where each Series describes the potential energy
+        gradient for the lambdas state that is perturbed.
+
```
orbeckst, line 116: You can use intersphinx links such as
```reST
.... :class:`~pandas.DataFrame` ... :class:`pandas.Series` ...
```
see the [python roles in sphinx](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects)

orbeckst (changes requested):
Please see comments – mainly concerns about testing/coverage and minor reST things.

Sorry for the brevity.

src/alchemlyb/estimators/ti_.py
xiki-tempula: I don't think that this will happen in the current setup. I removed this else block.

xiki-tempula:
@orbeckst Thank you for the review. I have adopted the changes.

orbeckst:
can you please merge master? thanks

xiki-tempula:
@orbeckst Thanks for the review. Change updated.

src/alchemlyb/estimators/ti_.py
orbeckst: @xiki-tempula can you please try rewriting without the `else` block?

orbeckst (changes requested):
See comment.

Also, is the function visible in the docs?

xiki-tempula:
@orbeckst Sorry. I see your point now. I thought that you were talking about other parts of the code. The else has been removed and deindented.

This function should be visible from this page https://alchemlyb.readthedocs.io/en/latest/estimators/alchemlyb.estimators.TI.html#alchemlyb.estimators.TI, when it is merged,

orbeckst (approved):

PR #114, opened 2021-02-28
Implement the workflow pipeline

xiki-tempula:
I will try to Implement the workflow pipeline as is discussed in #111 .

I noticed that the subsampling part will be refactored with #98. Given I cannot leave this part blank, I have made a minimal implementation of the corresponding option and I'm happy to change it when #98 is merged.

All the information is logged using logging #34.

PR #121, opened 2021-03-29
Plot convergence

xiki-tempula:
Plot Alchemical Analysis style convergence plot.


src/alchemlyb/visualisation/convergence.py
```diff
@@ -0,0 +1,72 @@
...
+    forward : List
+        A list of free energy estimate from the first X% of data.
+    forward_error : List
+        A list of error from the first X% of data.
+    backward : List
+        A list of free energy estimate from the last X% of data.
+    backward_error : List
+        A list of error from the last X% of data.
+    units : str
+        The label for the unit of the estimate. Default: :math:`kBT`
```
orbeckst, line 20: Do not use math mode for kBT. This is a Python string so the default should be what you can provide to the function, i.e., "kBT".

src/alchemlyb/visualisation/convergence.py
```diff
@@ -0,0 +1,72 @@
...
+                             zorder=1)
+    line1 = ax.errorbar(f_ts, forward, yerr=forward_error, color='#736AFF',
+                        lw=3, zorder=2, marker='o',
+                        mfc='w', mew=2.5, mec='#736AFF', ms=12,)
+    line2 = ax.errorbar(r_ts, backward, yerr=backward_error, color='#C11B17',
+                        lw=3, zorder=3, marker='o',
+                        mfc='w', mew=2.5, mec='#C11B17', ms=12, )
+
+    plt.xticks(r_ts[::2], fontsize=10)
+    plt.yticks(fontsize=10)
```
orbeckst, line 59: This is a purely nit-picky comment: If we're using the oo API for matplotlib then I'd use methods of `ax` instead of `plt.XXX`.

src/alchemlyb/visualisation/convergence.py
```diff
@@ -0,0 +1,72 @@
...
+
+    leg = plt.legend((line1[0], line2[0]), (r'$Forward$', r'$Reverse$'), loc=9,
+                    prop=FP(size=18), frameon=False)
+    plt.xlabel(r'$\mathrm{Fraction\/of\/the\/simulation\/time}$', fontsize=16,
+              color='#151B54')
```
orbeckst, line 64:
```suggestion
    plt.xlabel(r'Fraction of the simulation time', fontsize=16, color='#151B54')
```

src/alchemlyb/visualisation/convergence.py
```diff
@@ -0,0 +1,72 @@
...
+    plt.ylabel(r'$\mathrm{\Delta G\/%s}$' % units, fontsize=16,
+              color='#151B54')
```
orbeckst, line 66:
```suggestion
    plt.ylabel(r'$\Delta G$ (%s)'.format(units), fontsize=16, color='#151B54')
```

orbeckst: Manually check that this works.

docs/visualisation/alchemlyb.visualisation.plot_convergence.rst
```diff
@@ -0,0 +1,19 @@
...
+Plot the Forward and Backward Convergence
+=========================================
+
+The function :func:`~alchemlyb.visualisation.plot_convergence` allows
+the user to visualise the convergence by plotting the free energy change
+computed using the equilibrated snapshots between the proper target time frames
+in both forward (data points are stored in `forward` and `forward_error`) and
+reverse (data points are stored in `backward` and `backward_error`) directions.
+The unit in the y axis could be labelled to other units by setting *units*,
+which by default is :math:`kBT`. The user can pass :class:`matplotlib.axes.Axes` into
```
orbeckst, line 12:
```suggestion
which by default is :math:`k_B T` ("kT"). The user can pass :class:`matplotlib.axes.Axes` into
```

Can we make the string to indicate k_B T just "kT" instead of "kBT"? The latter puts too much emphasis on the "B" and for physicists at least, we're just talking about kT.

src/alchemlyb/visualisation/convergence.py
orbeckst: An can we use "kT" instead of "kBT"? (See comment elsewhere)

orbeckst: Add a note that setting the unit does not convert anything – this is purely for setting labels. The data already need to be in the desired unit.

orbeckst (changes requested):
Please see comments, thanks!!

src/alchemlyb/visualisation/convergence.py
xiki-tempula: @orbeckst Thank you for the suggestion. In https://github.com/alchemistry/alchemlyb/pull/114#discussion_r603577969 you mentioned that kT should be formatted as latex. I initially thought that it means using the math model but I guess this is not the case. I wonder if I can have an example of how this should be formatted? Thank you.

docs/visualisation/alchemlyb.visualisation.plot_convergence.rst
xiki-tempula: I have change this to :math:`kT`

src/alchemlyb/visualisation/convergence.py
orbeckst: I've been confused about what the argument does. Given that it's just a label, it should be clear that this is a Python string. You could set it to `r"$k_B R$"`, which does LaTeX formatting in matplotlib. It should always be clear that the argument is a string. 

Sorry if this was confusing.

docs/visualisation/alchemlyb.visualisation.plot_convergence.rst
xiki-tempula: @orbeckst Thank you for the explanation. In that case, I will change the doc for the function to 'kt' and the sphinx doc to :math:`kT` so that it is more clear.

src/alchemlyb/visualisation/convergence.py
xiki-tempula: @orbeckst Thank you for the suggestion. Just to confirm that you felt it is better to not label the energy unit in the Y-axis?

xiki-tempula:
@orbeckst Thanks for the review. So I have changed all the text to kT instead of kBT.
For the y label, currently, I will just use what is being inputted. 
Do you think it would be useful to just assume that the user input could be rendered correctly?

```
def plot_convergence(forward, forward_error, backward, backward_error,
                     units='kT', ax=None):
    ax.set_ylabel(r'$\Delta G$ (${}$)'.format(units), fontsize=16, color='#151B54')
```

Or allow the user to set their own latex input?

```
def plot_convergence(forward, forward_error, backward, backward_error,
                     units='$kT$', ax=None):
    ax.set_ylabel(r'$\Delta G$ ({})'.format(units), fontsize=16, color='#151B54')
```

orbeckst:
Let the user decide so that they can render their own LaTeX.

xiki-tempula:
@orbeckst Thank you for the suggestion. The current version would just use whatever the user inputs.
`ax.set_ylabel(r'$\Delta G$ ({})'.format(units), fontsize=16, color='#151B54')`

orbeckst:
please resolve the conflicts, thanks

xiki-tempula:
@orbeckst Thanks for the review. Conflict resolved.

orbeckst (approved):

PR #129, opened 2021-05-15
Add attr to the dataframe

### LINKED ISSUE #125: A unified way of unit conversion
Currently, the internal unit is always kt. However, when presenting the data, it might be desirable to present the data in kcal/mol or kj/mol. Thus, it would be a good idea to have a discussion on how to do the unit conversion.

Currently, the only place that requires unit conversion is the plotting functions and later on the workflow #114.

My proposal is that the plotting function will be taking a conversion factor that converts kt to the desired unit.

xiki-tempula:
Fix #125


xiki-tempula:
@orbeckst Sorry for disturbing you. I wonder if you mind have a look and give some advice, please? Thank you.

src/alchemlyb/estimators/mbar_.py
```diff
@@ -92,7 +92,10 @@ def fit(self, u_nk):
                               columns=self.states_,
                               index=self.states_) for i in out]
 
-        (self.delta_f_, self.d_delta_f_, self.theta_) = attrs 
+        (self.delta_f_, self.d_delta_f_, self.theta_) = attrs
```
orbeckst, line 95: Maybe we want to rename those `attrs` to something else to avoid confusion. `free_energy_differences` ?

src/alchemlyb/parsing/gmx.py
```diff
@@ -105,6 +105,9 @@ def extract_u_nk(xvg, T):
 
     u_k.name = 'u_nk'
 
+    u_k.attrs['temperature'] = T
+    u_k.attrs['energy_unit'] = 'kT'
+
```
orbeckst, line 110: We could wrap attribute setting so that we can use old pandas. 
```python
# remove once we can require pandas >= 1.2 (after Python 3.
if pd.__version__ >= "1.2":   # do this version check properly
   def pd_setattrs(df, data):
         df.attrs = data
else:
   def pd_setattrs(df, data):
        if not isinstance(data, dict):
              raise TypeError
        with warnings.catch_warnings():
              warnings.simplefilter("ignore")
              setattr(df, 'attrs', data) 
```
and then use `pd_setattrs()` everywhere where you set the attributes for the first time.

orbeckst: This could live in util or somewhere similar.

orbeckst (commented):
Initial comments
- make 2.7/3.5 removal separate PR
- need to keep 3.6 around (why does pd not publish packages for 3.6 ???????) so we need to work-around

src/alchemlyb/estimators/mbar_.py
xiki-tempula: This could be something that could be improved.
but delta_f_, d_delta_f_ are used in all estimators and on the pymbar module.

orbeckst: I meant the use of the name `attrs` on the RHS. Now that we are using `df.attrs`, it is very confusing because these free energy diffs aren't the attrs that we are using for units.

src/alchemlyb/parsing/gmx.py
xiki-tempula: @orbeckst Sorry, it is a bit unclear to me of what this function is trying to achieve. There should be no problem if one were to only use the alchemlyb functions when pandas < 1.2, but if one were to manipulate the data frame by themselves, pandas < 1.2 would then drop the attrs.

orbeckst:
@xiki-tempula I am a bit busy over the next few days, sorry if I don't get to reviewing anything. Please feel free to ping me around Wed/Thu again. Thank you for your understanding & patience. I really appreciate the work you're putting in here.

src/alchemlyb/parsing/gmx.py
orbeckst: When I tried to set `df.attrs`, pandas complained that I was trying to add a column but set the attribute nevertheless. My idea was to just filter the warnings.

It was not clear to me what the problem with attrs for 1.0 ≤ pandas < 1.2 was: Are you saying that only for ≥ 1.2, slicing a df will retain the attrs on the sliced copy?

src/alchemlyb/estimators/bar_.py
```diff
@@ -128,5 +128,7 @@ def fit(self, u_nk):
         self.d_delta_f_ = pd.DataFrame(np.sqrt(ad_delta + ad_delta.T),
                                        columns=self.states_,
                                        index=self.states_)
+        self.delta_f_.attrs = u_nk.attrs
+        self.d_delta_f_.attrs = u_nk.attrs
```
orbeckst, line 132: Does this raise a warning or is this find with pandas ≥ 1.0 ?

src/alchemlyb/estimators/mbar_.py
```diff
@@ -88,11 +88,14 @@ def fit(self, u_nk):
 
         # set attributes
         out = self._mbar.getFreeEnergyDifferences(return_theta=True)
-        attrs = [pd.DataFrame(i,
+        free_energy_differences = [pd.DataFrame(i,
```
orbeckst, line 91: indentation of following lines

src/alchemlyb/parsing/gmx.py
```diff
@@ -105,6 +105,9 @@ def extract_u_nk(xvg, T):
 
     u_k.name = 'u_nk'
 
+    u_k.attrs['temperature'] = T
+    u_k.attrs['energy_unit'] = 'kT'
```
orbeckst, line 109: Does this work without a warning with pandas ≥ 1.0?

src/alchemlyb/parsing/gmx.py
xiki-tempula: @orbeckst 

> When I tried to set df.attrs, pandas complained that I was trying to add a column but set the attribute nevertheless. My idea was to just filter the warnings.

For pandas == 1.0.5,
```
>>> d = {'col1': [1, 2], 'col2': [3, 4]}
>>> df = pd.DataFrame(data=d)
>>> df.attrs
{}
>>> df.attrs = {1:1}
```
yields no warning for me.

> It was not clear to me what the problem with attrs for 1.0 ≤ pandas < 1.2 was: Are you saying that only for ≥ 1.2, slicing a df will retain the attrs on the sliced copy?

Yes.

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Returns
+    -------
+    DataFrame
+        `df` converted.
+    """
+    assert 'temperature' in df.attrs, 'Attribute temperature not found in the' \
+                                      ' input Dataframe.'
+    assert 'energy_unit' in df.attrs, 'Attribute energy_unit not found in the' \
+                                      ' input Dataframe.'
```
orbeckst, line 27: Replace `assert` with raising a `TypeError`, saying that we need a df with attrs according to https://alchemlyb.readthedocs.io/en/stable/parsing.html#standard-forms-of-raw-data . Ultimately, the dfs are user input so we have to check them properly.

(`assert` can be optimized away and should not be used to check user input.)

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,91 @@
...
+        new_df.attrs['energy_unit'] = 'kT'
+        return new_df
+    elif df.attrs['energy_unit'] == 'kcal/mol':
+        new_df = df.copy()
+        new_df = new_df / (Boltzmann_constant * df.attrs['temperature'] *
+                           Avogadro_constant / 1000 * kJ2kcal)
+        new_df.attrs['energy_unit'] = 'kT'
+        return new_df
+    else:
+        raise NameError('energy_unit {} can only be kT, kJ/mol or ' \
```
orbeckst, line 43: This is a `ValueError` (wrong value).

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Returns
+    -------
+    DataFrame
+        `df` converted.
+    """
+    kt_df = to_kT(df)
+    kt_df = kt_df * (Boltzmann_constant * df.attrs['temperature'] *
+                           Avogadro_constant / 1000 * kJ2kcal)
+    return kt_df
```
orbeckst, line 64: set unit of `kt_df`

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Returns
+    -------
+    DataFrame
+        `df` converted.
+    """
+    kt_df = to_kT(df)
+    kt_df = kt_df * (Boltzmann_constant * df.attrs['temperature'] *
+                           Avogadro_constant / 1000)
+    return kt_df
```
orbeckst, line 83: set unit of `kt_df`

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,91 @@
...
+
+def pass_attrs(func):
+    '''Pass the attributes from the input dataframe to the output dataframe'''
+    def wrapper(input_dataframe, *args,**kwargs):
+        dataframe = func(input_dataframe, *args,**kwargs)
+        dataframe.attrs = input_dataframe.attrs
+        return dataframe
+    return wrapper
```
orbeckst, line 91: newline at end

orbeckst: This decorator should live elsewhere because it is used in different parts of the code and would also be useful for users.

For right now, can we just put it in the top-level `__init__.py` and you `from .. import pass_attrs`. Maybe we will later do a top-level utils.py.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -6,6 +6,7 @@
 from pymbar.timeseries import (statisticalInefficiency,
                                detectEquilibration,
                                subsampleCorrelatedData, )
+from ..postprocessors.units import pass_attrs
```
orbeckst, line 9: Switch to
```suggestion
from .. import pass_attrs
```

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,91 @@
...
+    DataFrame
+        `df` converted.
+    """
+    kt_df = to_kT(df)
+    kt_df = kt_df * (Boltzmann_constant * df.attrs['temperature'] *
+                           Avogadro_constant / 1000)
+    return kt_df
+
+def pass_attrs(func):
+    '''Pass the attributes from the input dataframe to the output dataframe'''
```
orbeckst, line 86: Add an example. Show the function signature of functions that it can be applied to, i.e., takes df as input first arg and returns a df.

src/alchemlyb/tests/test_units.py
```diff
@@ -0,0 +1,184 @@
...
+    assert u_nk.attrs['energy_unit'] == 'kT'
+
+def test_extract_dHdl():
+    '''Test if extract_u_nk assign the attr correctly'''
+    dataset = load_benzene()
+    dhdl = extract_dHdl(dataset['data']['Coulomb'][0], 310)
+    assert dhdl.attrs['temperature'] == 310
+    assert dhdl.attrs['energy_unit'] == 'kT'
+
+class Test_preprocessing():
```
orbeckst, line 30: add as `TestUnits` to the `test_preprocessors.py` so that in the future it is easier to add tests when anything changes in preprocessing.

src/alchemlyb/tests/test_units.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def test_equilibrium_detection(self, dhdl):
+        '''Test if extract_u_nk assign the attr correctly'''
+        dataset = load_benzene()
+        dhdl = extract_dHdl(dataset['data']['Coulomb'][0], 310)
+        new_dhdl = equilibrium_detection(dhdl)
+        assert new_dhdl.attrs['temperature'] == 310
+        assert new_dhdl.attrs['energy_unit'] == 'kT'
+
+class Test_estimator():
```
orbeckst, line 82: add as `TestUnits` to the `test_*_estimators.py` so that in the future it is easier to add tests when anything changes in estimators.

src/alchemlyb/tests/test_units.py
```diff
@@ -0,0 +1,184 @@
...
+        assert bar.d_delta_f_.attrs['energy_unit'] == 'kT'
+
+    def test_mbar(self, u_nk):
+        mbar = MBAR().fit(u_nk)
+        assert mbar.delta_f_.attrs['temperature'] == 300
+        assert mbar.delta_f_.attrs['energy_unit'] == 'kT'
+        assert mbar.d_delta_f_.attrs['temperature'] == 300
+        assert mbar.d_delta_f_.attrs['energy_unit'] == 'kT'
+
+class Test_visualisation():
```
orbeckst, line 132: add as `TestUnits` to the `test_visualisation.py` so that in the future it is easier to add tests when anything changes in viz.

src/alchemlyb/visualisation/dF_state.py
```diff
@@ -78,8 +79,19 @@ def plot_dF_state(estimators, labels=None, colors=None, units='kT',
...
+                    dF.append(to_kT(dhdl.delta_f_).iloc[i, i + 1])
+                    error.append(to_kT(dhdl.d_delta_f_).iloc[i, i + 1])
+                elif units == 'kJ/mol':
+                    dF.append(to_kJmol(dhdl.delta_f_).iloc[i, i + 1])
+                    error.append(to_kJmol(dhdl.d_delta_f_).iloc[i, i + 1])
+                elif units == 'kcal/mol':
+                    dF.append(to_kcalmol(dhdl.delta_f_).iloc[i, i + 1])
+                    error.append(to_kcalmol(dhdl.d_delta_f_).iloc[i, i + 1])
+                else:
+                    raise NameError('energy_unit {} can only be kT, kJ/mol '
```
orbeckst, line 92: ValueError – but isn't that already being raised with the conversion functions?

src/alchemlyb/visualisation/dF_state.py
```diff
@@ -78,8 +79,19 @@ def plot_dF_state(estimators, labels=None, colors=None, units='kT',
...
-                error.append(dhdl.d_delta_f_.iloc[i, i+1])
+                if units == 'kT':
+                    dF.append(to_kT(dhdl.delta_f_).iloc[i, i + 1])
+                    error.append(to_kT(dhdl.d_delta_f_).iloc[i, i + 1])
+                elif units == 'kJ/mol':
+                    dF.append(to_kJmol(dhdl.delta_f_).iloc[i, i + 1])
+                    error.append(to_kJmol(dhdl.d_delta_f_).iloc[i, i + 1])
+                elif units == 'kcal/mol':
+                    dF.append(to_kcalmol(dhdl.delta_f_).iloc[i, i + 1])
+                    error.append(to_kcalmol(dhdl.d_delta_f_).iloc[i, i + 1])
```
orbeckst, line 90: Too much code duplications. Use a dispatch table
```python
_converters = {'kT': to_kT, 'kJ/mol': to_kJmol, 'kcal/mol': to_kcalmol}
try:
   convert = _converters[units]
except KeyError:
   raise ValueError(f"Energy unit {units} is not supported, choose one of {list(_converters.keys())}")

dF.append(convert(dhdl.delta_f_).iloc[i, i + 1])
error.append(convert(dhdl.d_delta_f_).iloc[i, i + 1])
```

src/alchemlyb/visualisation/dF_state.py
orbeckst: If you use the dispatch table approach, it's already included.

src/alchemlyb/visualisation/ti_dhdl.py
```diff
@@ -60,6 +62,19 @@ def plot_ti_dhdl(dhdl_data, labels=None, colors=None, units='kT',
...
+    for dhdl in dhdl_list:
+        if units == 'kT':
+            new_unit.append(to_kT(dhdl))
+        elif units == 'kJ/mol':
+            new_unit.append(to_kJmol(dhdl))
+        elif units == 'kcal/mol':
+            new_unit.append(to_kcalmol(dhdl))
+        else:
+            raise NameError('energy_unit {} can only be kT, kJ/mol or ' \
+                            'kcal/mol.'.format(units))
```
orbeckst, line 76: use dispatch table

src/alchemlyb/visualisation/dF_state.py
orbeckst: You could put the dispatch table into `units` and import it.

orbeckst (changes requested):
Looking already quite good. I had various comments and one important additional request: We need to document this change in behavior under https://alchemlyb.readthedocs.io/en/stable/parsing.html#standard-forms-of-raw-data so please add text that explains how we annotate units and what _users_ have to do (namely, use `pass_attrs`.

Furthermore, add documentation about the new `postprocessing` module to the docs.

Update the visualization docs because now units are used for conversion.

Update CHANGES (these are big important ones). 

We might need another round of reviews but this is already looking really nice. Great work @xiki-tempula !

src/alchemlyb/estimators/bar_.py
xiki-tempula: @orbeckst There isn't any warning as https://travis-ci.org/github/alchemistry/alchemlyb/jobs/773410900 shows the py36 did not give more warning than py37.

src/alchemlyb/parsing/gmx.py
xiki-tempula: It works without any warning.

xiki-tempula:
@orbeckst Thanks for the review. The pandas attrs is a lot more experimental than I thought and I have added some more things to get it to work.

orbeckst:
Sorry, didn't have time to actually look at your code. Is `attr` stable in pandas ≥ 1.2 and do you only need workarounds for pandas 1.0 < 1.2?

xiki-tempula:
@orbeckst This is what I initially thought but it seems that this feature is a lot less mature than I thought.
For example, I need to have a custom concat (alchemlyb.concat) to replace pd.concat used in many docs.
However, the pass_attrs decorator should be able to correct for any loss of attrs during data frame manipulation.

orbeckst:
Random collection of approaches to the problem of df + units
- [astropy.tables](https://docs.astropy.org/en/stable/table/) (not a df but perhaps of interest)
- [pint-pandas](https://pint.readthedocs.io/en/0.16.1/pint-pandas.html) adds `pint` support to pandas (but is still experimental); I don't know what the performance implications are (and we originally wanted to keep alchemlyb lean on dependencies but something to investigate, especially if it were to simplify code/enable new things)
- build a custom DataFrame; perhaps use [pandas Extension Array](https://pandas.pydata.org/docs/reference/api/pandas.api.extensions.ExtensionArray.html) (pint uses EA)

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -191,3 +195,54 @@ class TestEquilibriumDetection(TestSlicing, CorrelatedPreprocessors):
...
+        assert new_dhdl.attrs['energy_unit'] == 'kT'
+
+    def test_kcal2kt(self, dhdl):
+        dhdl.attrs['energy_unit'] = 'kcal/mol'
+        new_dhdl = to_kT(dhdl)
+        assert new_dhdl.attrs['energy_unit'] == 'kT'
+
+    def test_unknown2kt(self, dhdl):
+        with pytest.raises(ValueError):
+            dhdl.attrs['energy_unit'] = 'ddd'
```
orbeckst, line 223: move this line outside the with block, keep it minimal

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -191,3 +195,54 @@ class TestEquilibriumDetection(TestSlicing, CorrelatedPreprocessors):
...
+            new_dhdl = to_kT(dhdl)
```
orbeckst, line 224: no need for assignment, or use `_ = ...` to indicate a throwaway value

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -191,3 +195,54 @@ class TestEquilibriumDetection(TestSlicing, CorrelatedPreprocessors):
...
+class Test_Units():
+    '''Test the preprocessing module.'''
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def dhdl():
+        dataset = load_benzene()
+        dhdl = extract_dHdl(dataset['data']['Coulomb'][0], 310)
+        return dhdl
+
+    def test_kt2kt(self, dhdl):
```
orbeckst, line 207: The tests for unit conversion should be with the postprocessing tests as these functions live there. Keep the tests localized with the code in the same module as much as possible.If someone implements new functions in postprocessing, they will immediately see the tests in test_postprocessing.py (but not in test_preprocessing.py).

src/alchemlyb/tests/test_units.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+import pandas as pd
+
+import alchemlyb
+from alchemtest.gmx import load_benzene
+from alchemlyb.parsing.gmx import extract_dHdl
+from alchemlyb.postprocessors.units import to_kT
+
```
orbeckst, line 8: Add an XFAIL test for using `pandas.concat`. When this test passes we know that pandas fixed attrs.

orbeckst: Add an XFAIL test for using pandas loc/iloc slicing. When this test passes we know that pandas fixed attrs.

src/alchemlyb/visualisation/dF_state.py
```diff
@@ -78,8 +76,18 @@ def plot_dF_state(estimators, labels=None, colors=None, units='kT',
         error = []
         for dhdl in dhdl_list:
             for i in range(len(dhdl.delta_f_) - 1):
-                dF.append(dhdl.delta_f_.iloc[i, i+1])
-                error.append(dhdl.d_delta_f_.iloc[i, i+1])
+                _converters = {'kT': to_kT, 'kJ/mol': to_kJmol,
+                               'kcal/mol': to_kcalmol}
```
orbeckst, line 80: Define the `_converters` _outside_ of the loop!

src/alchemlyb/visualisation/ti_dhdl.py
```diff
@@ -60,6 +59,20 @@ def plot_ti_dhdl(dhdl_data, labels=None, colors=None, units='kT',
         for dhdl in dhdl_data:
             dhdl_list.extend(dhdl.separate_dhdl())
 
+    # Convert unit
+    new_unit = []
+    for dhdl in dhdl_list:
+        _converters = {'kT': to_kT, 'kJ/mol': to_kJmol,
+                       'kcal/mol': to_kcalmol}
```
orbeckst, line 66: define outside of the loop ... or just define a function  it in `postprocessors.units`
```python
def get_unit_converter(unit):
   try:
       return _converters[unit]
   except KeyError:
       raise ValueError(...)
```

src/alchemlyb/visualisation/dF_state.py
orbeckst: Or better, use `get_unit_converter()` — see other comment.

docs/parsing.rst
```diff
@@ -94,10 +98,50 @@ For datasets that sample only a single :math:`\lambda` parameter, then the DataF
...
+.. autofunction:: alchemlyb.concat
+
+For more complex data manipulation functions, a decorator
+:func:`alchemlyb.pass_attrs` could be used to pass the metadata from the input
+data frame (first positional argument) to the output dataframe. ::
+
+    >>> from alchemlyb import pass_attrs
+    >>> @pass_attrs
+    >>> def slice(dataframe):
+    >>>     return dataframe.loc[::10]
```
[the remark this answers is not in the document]
xiki-tempula: slicing is safe with pandas > 1.2

xiki-tempula:
@orbeckst Thanks for the review. I will address the changes during the weekends.
With regard to `alchemlyb.concat` and `pd.concat`, so I will replace the code in the doc with `alchemlyb.concat` but leave `pd.concat` in the tests?

orbeckst:
Not quite. Use `alchemlyb.concat` in all the tests where they are now, as you have it in the PR. We want to test our code for correctness. But _add a few additional tests with `pd.concat`_ that will fail right now because attrs is not correctly propagated. Mark them with XFAIL. We just need one or two of these tests; they will let us know (by suddenly passing) when pandas can deal with passing attrs.

docs/parsing.rst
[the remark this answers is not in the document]
orbeckst: Ok, I think you convinced me to drop 3.6 as well. I'll open a separate issue so that other people can chime in if they want to.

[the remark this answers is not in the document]
orbeckst: I opened #140 to drop 3.6 and follow NEP29.

I admit, it took me a while to come around after having been opposed to it in #130 ...

orbeckst:
@xiki-tempula,  with PR #141 in the works, you can simplify this PR.
- remove Python 3.6 from the CI matrix
- require pandas >= 1.2
- remove any workarounds that would have beenn necessary for pandas < 1.2

src/alchemlyb/tests/test_units.py
xiki-tempula: @orbeckst loc/iloc slicing is safe now with pandas>=1.2

docs/parsing.rst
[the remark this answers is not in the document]
xiki-tempula: Since slicing is safe now, I have used concat as an example here.

xiki-tempula:
@orbeckst Thanks for the review and the advice. I think I have addressed all the changes. Would you mind have a look, please? Thank you.

orbeckst:
I won't be able to review until middle of the week, sorry. Thanks for your patience.

orbeckst:
Can you please rebase against master so that the new CI runs? Thanks.

xiki-tempula:
@orbeckst Thanks for the advice. I have merged the master to this branch.

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,122 @@
+'''Unit conversion and constants.'''
+
+from scipy.constants import R, calorie
+
+kJ2kcal = 1 / calorie
+R_kJmol = R / 1000
```
orbeckst, line 6: add doc as in
```suggestion
#: conversion factor from kJ to kcal, based on :data:`scipy.constants.calorie`
kJ2kcal = 1 / calorie

#: Gas constant :math:`R` in kJ/(mol K), based on :data:`scipy.constants.R`
R_kJmol = R / 1000
```

I like your approach here.
```

Comment (orbeckst) [reply src/alchemlyb/postprocessors/units.py:6]: Also add these constants to the documentation somewhere
```reST
.. autodata:: kJ2kcal
.. autodata:: R_kJmol
```

docs/postprocessors/alchemlyb.postprocessors.units.rst
```diff
@@ -0,0 +1,23 @@
...
+.. automodule:: alchemlyb.postprocessors.units
+
+.. autofunction:: to_kT
+
+.. autofunction:: to_kcalmol
+
+.. autofunction:: to_kJmol
+
+   
+
```
orbeckst, line 13: Also document the data and the `get_converter` function. Make a separate section "Constants and auxiliary functions" and put the real converters in a section "Unit converters".

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,122 @@
...
+    -------
+    DataFrame
+        `df` converted.
+    """
+    kt_df = to_kT(df, T)
+    kt_df *= R_kJmol * df.attrs['temperature']
+    kt_df.attrs['energy_unit'] = 'kJ/mol'
+    return kt_df
+
+def get_unit_converter(units):
```
orbeckst, line 101: make it appear in docs

src/alchemlyb/postprocessors/units.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Parameters
+    ----------
+    units : str
+        The unit that the function converts to.
+
+    Returns
+    -------
+    func
+        converter.
```
orbeckst, line 112: no period; maybe link to the docs that list the possible converter functions

orbeckst: add a versionadded

orbeckst (changes requested):
looks pretty good, some minor comments.

I didn't manage to look at everything in as much detail as I wanted, If I see more I add it.

xiki-tempula:
@orbeckst Thanks for the review. I have updated the docs and checked it.

src/alchemlyb/__init__.py
```diff
@@ -1,4 +1,45 @@
...
+    -------
+    DataFrame
+        Concatenated pandas object.
+
+    Raises
+    ------
+    ValueError
+        If not all pandas objects have the same attrs.
+ 
+
```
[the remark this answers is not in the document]
orbeckst: Check if this renders correctly. If we don't have pandas in the intersphinx mapping, add it to sphinx conf.py.

Also make sure that there are at least two empty lines before the version added.

CHANGES
```diff
@@ -18,12 +18,18 @@ The rules for this file:
...
+  #129).
+  - The parser output dataframe with metadata (PR #129).
 
 Deprecations
 
 Fixes
 
 Changes
+  - alchemlyb.concat added to replace pd.concat (PR #129).
+  - postprocessors.units module for unit conversion (PR #129).
```
orbeckst, line 31: Add under enhancements. (The "Changes" parts are listed separately, see other comments.)

Reference #125 (generally, the issue is more important than the PR, you can include both if you like, but definitely reference the issue)

CHANGES
```diff
@@ -18,12 +18,18 @@ The rules for this file:
   * 0.5.0
 
 Enhancements
+  - Visualisation module will change the data according to input unit (PR
```
orbeckst, line 21: Reference #125 (generally, the issue is more important than the PR, you can include both if you like, but definitely reference the issue)

CHANGES
```diff
@@ -18,12 +18,18 @@ The rules for this file:
...
+  #129).
+  - The parser output dataframe with metadata (PR #129).
```
orbeckst, line 23: I would list this under Changes because it changes our data format.

docs/estimators-fep.rst
```diff
@@ -16,8 +16,8 @@ and parse the datafiles separately for each alchemical leg using :func:`alchemly
     >>> from alchemlyb.parsing.gmx import extract_u_nk
     >>> import pandas as pd
 
-    >>> u_nk_coul = pd.concat([extract_u_nk(xvg, T=300) for xvg in bz['Coulomb']])
-    >>> u_nk_vdw = pd.concat([extract_u_nk(xvg, T=300) for xvg in bz['VDW']])
+    >>> u_nk_coul = alchemlyb.concat([extract_u_nk(xvg, T=300) for xvg in bz['Coulomb']])
+    >>> u_nk_vdw = alchemlyb.concat([extract_u_nk(xvg, T=300) for xvg in bz['VDW']])
```
orbeckst, line 20: I assume you did a `git grep concat` and looked at every occurrence in the code?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -117,12 +117,12 @@ def test_u_nk_with_total_energy():
     dataset = load_water_particle_with_total_energy()
 
     # Check if the sum of values on the diagonal has the correct value
-    assert_almost_equal(_diag_sum(dataset), 47611377946.58586, decimal=4)
+    assert_almost_equal(_diag_sum(dataset), 47611374980.34574, decimal=4)
```
orbeckst, line 120: Why did the value change? Did we fix something?

Specifically, how did the new unit system change anything? We will need to be very clear in CHANGES that results will change. By how much?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -117,12 +117,12 @@ def test_u_nk_with_total_energy():
...
 
     # Check one specific value in the dataframe
     assert_almost_equal(
         extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0],
-        -11211.578357345974,
+        -11211.577658852531,
```
orbeckst, line 125: Why different?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -135,12 +135,12 @@ def test_u_nk_with_potential_energy():
     dataset = load_water_particle_with_potential_energy()
 
     # Check if the sum of values on the diagonal has the correct value
-    assert_almost_equal(_diag_sum(dataset), 16674041445589.646, decimal=2)
+    assert_almost_equal(_diag_sum(dataset), 16674040406778.867, decimal=2)
```
orbeckst, line 138: Why different?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -135,12 +135,12 @@ def test_u_nk_with_potential_energy():
...
 
     # Check one specific value in the dataframe
     assert_almost_equal(
         extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0],
-        -15656.558227621246,
+        -15656.557252200757,
```
orbeckst, line 143: Why different?

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -154,7 +154,7 @@ def test_u_nk_without_energy():
     dataset = load_water_particle_without_energy()
 
     # Check if the sum of values on the diagonal has the correct value
-    assert_almost_equal(_diag_sum(dataset), 20572988148877.555, decimal=2)
+    assert_almost_equal(_diag_sum(dataset), 20572986867158.184, decimal=2)
```
orbeckst, line 157: Why different?

src/alchemlyb/tests/test_units.py
```diff
@@ -0,0 +1,113 @@
...
+        assert 21.0 == pytest.approx(new_dhdl.iloc[0, 0], 0.1)
+
+    def test_unknown2kt(self, dhdl):
+        dhdl.attrs['energy_unit'] = 'ddd'
+        with pytest.raises(ValueError):
+            to_kT(dhdl)
+
+@pytest.mark.xfail
+def test_pd_concat():
+    '''Test if concat will preserve the metadata.'''
```
orbeckst, line 87: add a comment referencing the pandas issue; note that once this tests passes, we will increase the minimal pandas version to the version of pandas that passes this test and then remove `alchemlyb.concat`. 

Think about future developers who might not understand why we're testing a pandas feature. Tell them why.

src/alchemlyb/tests/test_visualisation.py
```diff
@@ -34,7 +35,9 @@ def test_plot_mbar_omatrix():
 def test_plot_ti_dhdl():
     '''Just test if the plot runs'''
     bz = load_benzene().data
-    dHdl_coul = pd.concat([extract_dHdl(xvg, T=300) for xvg in bz['Coulomb']])
+    dHdl_coul = alchemlyb.concat([extract_dHdl(xvg, T=300) for xvg in bz['Coulomb']])
+    dHdl_coul.attrs = extract_dHdl(load_benzene().data['Coulomb'][0],
+                                   T=300).attrs
```
orbeckst, line 40: Why is the explicit attr assignment needed here? I though `extract_dHdl` and `alchemlyb.concat()` were automatically handling this case??

src/alchemlyb/tests/test_visualisation.py
```diff
@@ -46,24 +49,33 @@ def test_plot_ti_dhdl():
                matplotlib.axes.Axes)
     assert isinstance(plot_ti_dhdl(ti_coul, labels=['Coul'], colors=['r']),
                matplotlib.axes.Axes)
-    dHdl_vdw = pd.concat([extract_dHdl(xvg, T=300) for xvg in bz['VDW']])
+    dHdl_vdw = alchemlyb.concat([extract_dHdl(xvg, T=300) for xvg in bz['VDW']])
+    dHdl_vdw.attrs = extract_dHdl(load_benzene().data['Coulomb'][0],
+                                   T=300).attrs
```
orbeckst, line 54: Why explicit attr assignment?

src/alchemlyb/tests/test_visualisation.py
```diff
@@ -123,3 +135,55 @@ def test_plot_convergence():
...
         matplotlib.axes.Axes)
 
+class Test_Units():
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def estimaters():
+        bz = load_benzene().data
+        dHdl_coul = alchemlyb.concat(
+            [extract_dHdl(xvg, T=300) for xvg in bz['Coulomb']])
+        dHdl_coul.attrs = extract_dHdl(load_benzene().data['Coulomb'][0], T=300).attrs
```
orbeckst, line 145: Why attr assignment?

src/alchemlyb/tests/test_visualisation.py
```diff
@@ -123,3 +135,55 @@ def test_plot_convergence():
...
+        ti = TI().fit(dHdl_coul)
+
+        u_nk_coul = alchemlyb.concat(
+            [extract_u_nk(xvg, T=300) for xvg in bz['Coulomb']])
+        u_nk_coul.attrs = extract_dHdl(load_benzene().data['Coulomb'][0], T=300).attrs
```
orbeckst, line 150: Why attr assignment?

src/alchemlyb/visualisation/convergence.py
```diff
@@ -33,7 +33,7 @@ def plot_convergence(forward, forward_error, backward, backward_error,
     `Alchemical Analysis <https://github.com/MobleyLab/alchemical-analysis>`_.
 
     The units variable is for labelling only. Changing it doesn't change the
-    unit of the underlying variable, which is in the unit of :math:`kT`.
+    unit of the underlying variable.
```
orbeckst, line 36: Does `plot_convergence()`  not do unit conversion, even though the other plotting functions do? 

Can you change it, not the least the CHANGES says that this is now working for all visualization functions.

orbeckst: Or remove `units` and take the label from the df `attr` — but having the same behavior as the other plotting functions would be much better and more consistent.

src/alchemlyb/visualisation/dF_state.py
```diff
@@ -46,9 +47,6 @@ def plot_dF_state(estimators, labels=None, colors=None, units='kT',
     The code is taken and modified from
     `Alchemical Analysis <https://github.com/MobleyLab/alchemical-analysis>`_.
 
-    The units variable is for labelling only. Changing it doesn't change the
-    unit of the underlying variable, which is in the unit of :math:`kT`.
-
```
orbeckst, line 51: add a versionchanged for the new behavior of `units`.

src/alchemlyb/visualisation/ti_dhdl.py
```diff
@@ -43,9 +45,6 @@ def plot_ti_dhdl(dhdl_data, labels=None, colors=None, units='kT',
     The code is taken and modified from
     `Alchemical Analysis <https://github.com/MobleyLab/alchemical-analysis>`_.
 
-    The units variable is for labelling only. Changing it doesn't change the
```
orbeckst, line 46: add a versionchanged for the new behavior of `units`.

orbeckst (changes requested):
Second part of review.
* Major question: Why do the GMX tests now produce different answers? This is a serious change and we need to be _very_ clear about what changed, why, and how much of a difference it makes — _if_ the change is necessary and not a bug.
* There are tests where `attr` is explicitly assigned after use of `extract_dHdl` and `alchemlyb.concat()` and I don't understand why: I thought that alchemlyb will seamlessly handle units when using only alchemlyb functions.
* Mostly minor doc issues.
* Add `postprocessing` to [docs/api_principles.rst](https://github.com/alchemistry/alchemlyb/blob/master/docs/api_principles.rst)

src/alchemlyb/visualisation/convergence.py
xiki-tempula: There is currently no alchemlyb function that will generate the input for plot_convergence.
The user needs to generate the input file, which could be a simple python list.

docs/estimators-fep.rst
xiki-tempula: Yes.

src/alchemlyb/tests/test_visualisation.py
xiki-tempula: Sorry, I initially did this and then I felt it might be more useful to have alchemlyb.concat() and then I change pd.concat to alchemlyb.concat without removing this.

xiki-tempula:
@orbeckst Thank you for the review. I will do the changes when I got time.

> Major question: Why do the GMX tests now produce different answers? This is a serious change and we need to be very clear about what changed, why, and how much of a difference it makes — if the change is necessary and not a bug.

The original parser uses the gromacs constant https://manual.gromacs.org/documentation/2019/reference-manual/definitions.html

> R | gas constant | 8.3144621×10−3 kJ mol−1 K−1

The new R uses the scipy constant.

```
>>> from scipy.constants import R
>>> R
8.314462618
```

They differ by `(8.3144621-8.314462618)/8.3144621 = -6.230108382200071e-08`.

> There are tests where attr is explicitly assigned after use of extract_dHdl and alchemlyb.concat() and I don't understand why: I thought that alchemlyb will seamlessly handle units when using only alchemlyb functions.

Sorry, I initially did this when we only have pd.concat.
Then I felt it might be more useful to have alchemlyb.concat() instead of doing this everywhere, so I made it and change pd.concat to alchemlyb.concat. But I forget to remove this.

orbeckst:
Thanks for the explanation regarding R. Given that we can consistently convert back to the actual kJ/mol units that GROMACS provides, the change in the reference values in the tests is ok.

Let me know when the PR is ready for final review.

orbeckst:
We will need to add a note to CHANGES and a Warning to the gmx parser page, though, because the results in kT that people produced previously will now be slightly off.

src/alchemlyb/__init__.py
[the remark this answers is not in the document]
xiki-tempula: Yes, it is rendered correctly.

xiki-tempula:
@orbeckst Thanks for the review. I have addressed all the changes.
However, there seems to be a problem. I think I have configured the sphinx/conf.py correctly. 
But :attr:`pandas.DataFrame.attrs` and :attr:`scipy.constants.gas_constant` cannot be reference correctly.

CHANGES
```diff
@@ -18,12 +18,20 @@ The rules for this file:
...
+  - pass_attrs decorator for safely passing variable (issue #125, PR #129).
+  - alchemlyb.concat added to replace pd.concat (issue #125, PR #129).
 
 Deprecations
 
 Fixes
 
 Changes
+  - The scipy constant for gas constant is used for parsers instead of the
+  contestants used by the corresponding MD engine (issue #125, PR #129).
```
orbeckst, line 31:
```suggestion
    constants used by the corresponding MD engine (issue #125, PR #129).
    This leads to small changes in results from GROMACS simulations compared to the 
    previous releases.
```

src/alchemlyb/parsing/amber.py
```diff
@@ -273,6 +271,11 @@ def extract_u_nk(outfile, T):
     u_nk : DataFrame
         Reduced potential for each alchemical state (k) for each frame (n).
 
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 277:
```suggestion
        the constants used by the corresponding MD engine.
```

src/alchemlyb/parsing/amber.py
```diff
@@ -327,6 +330,12 @@ def extract_dHdl(outfile, T):
     -------
     dH/dl : Series
         dH/dl as a function of time for this lambda window.
+
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 337:
```suggestion
        the constants used by the corresponding MD engine.
```

src/alchemlyb/parsing/gmx.py
```diff
@@ -29,6 +26,11 @@ def extract_u_nk(xvg, T):
     u_nk : DataFrame
         Potential energy for each alchemical state (k) for each frame (n).
 
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 32:
```suggestion
        the constants used by the corresponding MD engine.        
        This leads to slightly different results for GROMACS input compared to 
        previous versions of alchemlyb.
```

src/alchemlyb/parsing/gmx.py
```diff
@@ -123,6 +125,11 @@ def extract_dHdl(xvg, T):
     dH/dl : Series
         dH/dl as a function of time for this lambda window.
 
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 131:
```suggestion
        the constants used by the corresponding MD engine.
        This leads to slightly different results for GROMACS input compared to 
        previous versions of alchemlyb.        
```

src/alchemlyb/parsing/gomc.py
```diff
@@ -28,6 +25,11 @@ def extract_u_nk(filename, T):
     u_nk : DataFrame
         Potential energy for each alchemical state (k) for each frame (n).
 
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 31:
```suggestion
        the constants used by the corresponding MD engine.
```

src/alchemlyb/parsing/gomc.py
```diff
@@ -104,6 +106,11 @@ def extract_dHdl(filename, T):
     dH/dl : Series
         dH/dl as a function of step for this lambda window.
 
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 112:
```suggestion
        the constants used by the corresponding MD engine.
```

src/alchemlyb/parsing/namd.py
```diff
@@ -27,6 +25,11 @@ def extract_u_nk(fep_file, T):
     u_nk : DataFrame
         Potential energy for each alchemical state (k) for each frame (n).
 
+
+    .. versionchanged:: 0.5.0
+        The :mod:`scipy.constants` is used for parsers instead of
+        the contestants used by the corresponding MD engine.
```
orbeckst, line 31:
```suggestion
        the constants used by the corresponding MD engine.
```

src/alchemlyb/parsing/gmx.py
```diff
@@ -29,6 +26,11 @@ def extract_u_nk(xvg, T):
     u_nk : DataFrame
         Potential energy for each alchemical state (k) for each frame (n).
 
+
```
orbeckst, line 29:
```suggestion

    Note
    -----
    Previous versions of alchemlyb (<0.5.0) used the `GROMACS value of the molar gas constant
    <https://manual.gromacs.org/documentation/2019/reference-manual/definitions.html>`_
    of  :math:`R = 8.3144621 \times 10^{−3}\, \text{kJ}\cdot\text{mol}^{-1}\cdot\text{K}^{-1}`
    instead of the scipy value :data:`scipy.constants.R` (see :mod:`alchemlyb.postprocessors.units`).
    The relative difference between the two values is :math:`6 \times 10^{-8}`. 
    
    Therefore, results in :math:`kT` for GROMACS data will differ between alchemlyb ≥0.5.0 and
    previous versions; the difference is on the order of :math:`10^{-7}` for typical cases.
    
    
```

src/alchemlyb/parsing/gmx.py
```diff
@@ -123,6 +125,11 @@ def extract_dHdl(xvg, T):
     dH/dl : Series
         dH/dl as a function of time for this lambda window.
 
+
```
orbeckst, line 128:
```suggestion
   
    Note
    -----
    Previous versions of alchemlyb (<0.5.0) used the `GROMACS value of the molar gas constant
    <https://manual.gromacs.org/documentation/2019/reference-manual/definitions.html>`_
    of  :math:`R = 8.3144621 \times 10^{−3}\, \text{kJ}\cdot\text{mol}^{-1}\cdot\text{K}^{-1}`
    instead of the scipy value :data:`scipy.constants.R` (see :mod:`alchemlyb.postprocessors.units`).
    The relative difference between the two values is :math:`6 \times 10^{-8}`. 
    
    Therefore, results in :math:`kT` for GROMACS data will differ between alchemlyb ≥0.5.0 and
    previous versions; the difference is on the order of :math:`10^{-7}` for typical cases.
    
    
```

docs/conf.py
```diff
@@ -174,7 +174,9 @@
 intersphinx_mapping = {'https://docs.python.org/': None,
                        'pandas': ('https://pandas.pydata.org/pandas-docs/stable/', None),
                        'pymbar': ('https://pymbar.readthedocs.io/en/latest/', None),
-                       'alchemtest': ('https://alchemtest.readthedocs.io/en/latest/', None)}
+                       'alchemtest': (
+                           'https://alchemtest.readthedocs.io/en/latest/', None),
+                       'scipy': ('http://docs.scipy.org/doc/scipy/reference/', None)}
```
orbeckst, line 179:
```suggestion
                       'scipy': ('https://docs.scipy.org/doc/scipy/reference/', None)}
```

orbeckst (approved):
Nice work, big step forward.

I have added minor corrections, which I will add in batch.

PR #135, opened 2021-06-04
Support for Interleaved Double-Wide Sampling files from NAMD

jhenin:


jhenin:
Well, looks like I broke the tests... I will look into this and update the PR.

jhenin:
I'm going to submit a new dataset to alchemtest for this.

orbeckst:
@jhenin please submit a test set PR for alchemtest and ping me so that I can quickly review.

orbeckst:
(sorry, didn't mean to be pushy, just wanted to say that _once_ you get to making a PR I am happy to quickly review it — just ping me to get my attention, please)

jhenin:
@orbeckst No worries. This update should do the trick.

src/alchemlyb/parsing/namd.py
```diff
@@ -33,61 +33,99 @@ def extract_u_nk(fep_file, T):
...
+                    lambda1 = "{0:.2f}".format(float(l[7]))
+                    lambda2 = "{0:.2f}".format(float(l[8]))
+
+                    # convert last window's work and times values to np arrays
+                    win_de_arr = beta * np.asarray(win_de)
+                    win_ts_arr = np.asarray(win_ts)
+
+                    if lambda_idws is not None:
+                    # Mimic classic DWS data
+                    # Arbitrarily match up fwd and bwd comparison energies on the same times
```
orbeckst, line 77: Document in a Note section that the parser may discard data.

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -5,6 +5,8 @@
 
 from alchemlyb.parsing.namd import extract_u_nk
 from alchemtest.namd import load_tyr2ala
+from alchemtest.namd import load_idws
```
orbeckst, line 8: This dataset only contains a single file; to test the functionality of parsing multiple files we should have an appropriate data set (or if it makes sense), double the existing data to fake it (maybe shuffle it?)

src/alchemlyb/parsing/namd.py
```diff
@@ -12,13 +12,13 @@
 k_b = 1.9872041e-3
 
 
-def extract_u_nk(fep_file, T):
-    """Return reduced potentials `u_nk` from NAMD fepout file.
+def extract_u_nk(fep_files, T):
+    """Return reduced potentials `u_nk` from NAMD fepout file(s).
```
orbeckst, line 16: Document that IDWS is now supported (reference to paper/NAMD docs?).

Document that parsing IDWS file can discard data.

Explain the use of the multiple `fep_files` as input. I am not familiar with NAMD's FEP code and output so I have no idea if this is obvious or not. A simple example might help and really _any_ docs are better than none.

orbeckst (changes requested):
Thanks @jhenin . I have an initial set of comments/questions.

Please also add an entry to CHANGES under 0.5.0 (your GitHub name and support for NAMD IDWS under Enhancements). 

Thank you!

src/alchemlyb/parsing/namd.py
orbeckst: Please also add a 
```reST


.. versionchanged:: 0.5.0
   Support for Interleaved Double-Wide Sampling files added. 
   `fep_files` can now be a list of filenames.
```

src/alchemlyb/parsing/namd.py
```diff
@@ -33,61 +33,99 @@ def extract_u_nk(fep_file, T):
...
 
     # create dataframe for results
     u_nk = pd.DataFrame(columns=['time','fep-lambda'])
 
     # boolean flag to parse data after equil time
     parsing = False
+    lambda_idws = None
 
+    if type(fep_files) is str:
+        fep_files = [fep_files]
```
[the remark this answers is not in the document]
jhenin: You're making it [difficult for me to cut corners](https://www.reddit.com/r/ProgrammerHumor/comments/o0f5b5/wait_youre_reviewing_before_approving_right/). I will change the data set so it's split into several files. Deep inside, I knew I really should have done that all along.

orbeckst:
We just switched the CI to GH actions so I needed to merge master and kick it... should report status in a few minutes (and I also fixed a bug in alchemtest, which I think _I_ introduced with my suggestion on your PR. Sorry about that.)

orbeckst:
I don’t know why codecov fails during the upload stage. We just changed the CI in #139 so there might be teething problems — maybe @dotsdl @xiki-tempula have any ideas?

xiki-tempula:
I will have a look

orbeckst:
Oops, sorry, codecov works just fine on the PR. I got confused with a codecov failure on @jhenin ‘s fork that I got a notification for. https://github.com/jhenin/alchemlyb/actions/runs/946402101  — please ignore me… it’s another 117F day and my brain is already mushy. (At least that’s my excuse.)

orbeckst:
Sorry, I only just saw that the GH workflows were waiting for admin approval — that's a new feature on GitHub to crack down on abuse of CI for bitcoin mining etc. Please don't hesitate to ping me if you need my attention.

orbeckst:
I resolved the conflict and update CHANGES (including your and @ttjoseph 's contribution). This started the CI, too.

orbeckst:
@jhenin and @ttjoseph please add yourselves to [AUTHORS](https://github.com/alchemistry/alchemlyb/blob/master/AUTHORS).

CHANGES
```diff
@@ -13,6 +13,14 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
+??/??/2021 jhenin, ttjoseph
```
orbeckst, line 16: Also add yourselves to AUTHORS

src/alchemlyb/parsing/namd.py
```diff
@@ -9,14 +9,71 @@
 
 k_b = R_kJmol * kJ2kcal
 
+def get_lambdas(fep_files):
```
orbeckst, line 12: Does this show up in the docs?

Build them locally with
```
python setup.py build_sphinx
```
and check.

(For some reason the RTD PR integration does not work #156.)

src/alchemlyb/parsing/namd.py
orbeckst: Please add the versionchanged to the docs, just updated to **0.6.0**.

src/alchemlyb/parsing/namd.py
orbeckst: Are users supposed to use `get_lambdas()` by themselves or would you consider this a "private" function of the module. If the latter, prefix with underscore and call it `_get_lambdas()` and then we also don't have to worry about documenting for users and we don't have to maintain the interface rigorously.

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +82,136 @@ def extract_u_nk(fep_file, T):
     u_nk : DataFrame
         Potential energy for each alchemical state (k) for each frame (n).
 
-
-    .. versionchanged:: 0.5.0
-        The :mod:`scipy.constants` is used for parsers instead of
-        the constants used by the corresponding MD engine.
-
```
orbeckst, line 32: Was there a reason to delete this?

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +82,136 @@ def extract_u_nk(fep_file, T):
...
+                        lambda_idws = None
+
+                    # If the lambdas are not what we thought they would be, return None, ensuring the calculation
+                    # fails. This can happen if fepouts where one window spans multiple fepouts are processed out of order
+                    if lambda1_at_start is not None \
+                        and (lambda1, lambda2, lambda_idws) != (lambda1_at_start, lambda2_at_start, lambda_idws_at_start):
+                        print("namd.py: extract_u_nk: Error: Lambdas changed unexpectedly while processing", fep_file)
+                        print(f"namd.py: extract_u_nk: Error: l1, l2, lidws: {lambda1_at_start}, {lambda2_at_start}, {lambda_idws_at_start} changed to {lambda1}, {lambda2}, {lambda_idws}")
+                        print(f"namd.py: extract_u_nk: Error: fep_file = {fep_file}; has_idws = {has_idws}")
+                        return None
```
orbeckst, line 152: Needs to be tested, check the coverage report (maybe the link https://app.codecov.io/gh/alchemistry/alchemlyb/compare/135/diff#diff-c3JjL2FsY2hlbWx5Yi9wYXJzaW5nL25hbWQucHk= works)

src/alchemlyb/parsing/namd.py
```diff
@@ -9,14 +9,71 @@
...
+                # We might not have a #NEW line so make the best guess
+                if l[0] == '#NEW':
+                    lambda1, lambda2 = float(l[6]), float(l[8])
+                    lambda_idws = float(l[10]) if 'LAMBDA_IDWS' in l else None
+                elif l[0] == '#Free':
+                    lambda1, lambda2, lambda_idws = float(l[7]), float(l[8]), None
+
+                # Make sure the lambda2 values are consistent
+                if lambda1 in lambda_fwd_map and lambda_fwd_map[lambda1] != lambda2:
+                    print(f'namd.py: get_lambdas: Error: fwd: lambda1 {lambda1} has lambda2 {lambda_fwd_map[lambda1]} but it should be {lambda2}')
```
orbeckst, line 49: Do not use `print()` in library code. Use a logger for *alchemlyb.parsers.NAMD* (see what the Amber parser does https://github.com/alchemistry/alchemlyb/blob/b068776354055115da6d52af24664ba3938985cc/src/alchemlyb/parsing/amber.py#L21 ) and issue a `warnings.warn` if necessary. 

If this is really an error then shouldn't you raise an exception such as `ValueError`?

src/alchemlyb/parsing/namd.py
```diff
@@ -9,14 +9,71 @@
...
+                    return None
+
+                lambda_fwd_map[lambda1] = lambda2
+
+                # Make sure the lambda_idws values are consistent
+                if lambda_idws is not None:
+                    if lambda1 in lambda_bwd_map and lambda_bwd_map[lambda1] != lambda_idws:
+                        print(f'namd.py: get_lambdas: Error: bwd: lambda1 {lambda1} has lambda_idws {lambda_bwd_map[lambda1]} but it should be {lambda_idws}')
```
orbeckst, line 57: Use a logger instead of `print` and shouldn't this raise a `ValueError`, too?

Fail early!

src/alchemlyb/parsing/namd.py
```diff
@@ -9,14 +9,71 @@
...
+                        return None
```
orbeckst, line 58: My concern is that returning `None` will silently discard data. It's cleaner to fail and have the user fix things.

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +82,136 @@ def extract_u_nk(fep_file, T):
...
+                    else:
+                        lambda_idws = None
+
+                    # If the lambdas are not what we thought they would be, return None, ensuring the calculation
+                    # fails. This can happen if fepouts where one window spans multiple fepouts are processed out of order
+                    if lambda1_at_start is not None \
+                        and (lambda1, lambda2, lambda_idws) != (lambda1_at_start, lambda2_at_start, lambda_idws_at_start):
+                        print("namd.py: extract_u_nk: Error: Lambdas changed unexpectedly while processing", fep_file)
+                        print(f"namd.py: extract_u_nk: Error: l1, l2, lidws: {lambda1_at_start}, {lambda2_at_start}, {lambda_idws_at_start} changed to {lambda1}, {lambda2}, {lambda_idws}")
+                        print(f"namd.py: extract_u_nk: Error: fep_file = {fep_file}; has_idws = {has_idws}")
```
orbeckst, line 151: Do not use `print()` in library code. Use a logger for *alchemlyb.parsers.NAMD* (see what the Amber parser does https://github.com/alchemistry/alchemlyb/blob/b068776354055115da6d52af24664ba3938985cc/src/alchemlyb/parsing/amber.py#L21 ).

Instead of returning `None`, raise the exception here; `ValueError` could be appropriate unless you have another idea.

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +82,136 @@ def extract_u_nk(fep_file, T):
...
+                    elif l[0] == 'FepE_back:':
+                        win_de_back.append(float(l[6]))
+                        win_ts_back.append(float(l[1]))
+
+                # Turn parsing on after line 'STARTING COLLECTION OF ENSEMBLE AVERAGE'
+                if '#STARTING' in l:
+                    parsing = True
+
+    if len(win_de) != 0 or len(win_de_back) != 0:
+        print('Warning: trailing data without footer line (\"#Free energy...\"). Interrupted run?')
```
orbeckst, line 206: Do not use `print()` in library code. Use a logger for *alchemlyb.parsers.NAMD* (see what the Amber parser does https://github.com/alchemistry/alchemlyb/blob/b068776354055115da6d52af24664ba3938985cc/src/alchemlyb/parsing/amber.py#L21 ) and issue a `warnings.warn` .

orbeckst (changes requested):
Thanks for the update, including the better test files. My major issues are (see comments)

* Replace any `print()` with a logger call (and a `warnings.warn()` for cases when logging is not enabled). `print` in library code is really annoying because you cannot nicely control the output from code that integrates the library.
* Instead of returning `None` and making calling code choke on the return value, raise an exception right in the code where it happens. I _assume_ that this is intended behavior (instead of having other code that works around the `None`). In any case, **fail early and clearly** and **don't guess** are principles that make code simpler and more robust, which is what alchemlyb is going for.
* The test coverage needs to be increased. Basically, we want tests for any of the exceptions and corner cases that you are looking for. This is important to reduce the maintenance burden and code-debt down in the future (and to just make sure that the code really does what it was intended to do). You can use `mock` or generate a bad input file on the fly in the tests.

src/alchemlyb/parsing/namd.py
ttjoseph: @orbeckst That link gives me an "GitHub API rate limit error" message. I'm unfamiliar with Codecov, but I can at least add a test for this

orbeckst: Hm, yes, I get that, too, when I am not logged in with my GitHub account.

Look at the running tests and then click on the codecov actions (e.g., https://github.com/alchemistry/alchemlyb/pull/135/checks?check_run_id=3314785981 ) and then go from there?

orbeckst: This leads me to https://app.codecov.io/gh/alchemistry/alchemlyb/compare/135/diff — when I tried this link in an anonynmous browser window without being logged in anywhere I could see everything.

src/alchemlyb/parsing/namd.py
ttjoseph: get_lambdas() shouldn't be exposed to users. Our next PR will have the underscore prefix.

src/alchemlyb/parsing/namd.py
ttjoseph: Thanks. I am not familiar with Codecov so I'll need to figure out how to tell it what my still-to-be-written new tests cover.

orbeckst:
Apologies, I hadn't seen that I needed to approve running of the CI. Please ping me when something like this is needed.

I merged PR alchemistry/alchemtest#55 so the dataset should now be available.

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +86,147 @@ def extract_u_nk(fep_file, T):
...
+    Note
+    ----
+    If the number of forward and backward samples in a given window are different,
+    the extra sample(s) will be discarded. This is typically zero or one sample.
 
     .. versionchanged:: 0.5.0
         The :mod:`scipy.constants` is used for parsers instead of
         the constants used by the corresponding MD engine.
 
+        Support for Interleaved Double-Wide Sampling files added.
```
orbeckst, line 98: Need new version here:
```suggestion
    .. versionchanged:: 0.6.0
        Support for Interleaved Double-Wide Sampling files added.
```

src/alchemlyb/parsing/namd.py
orbeckst: I am adding it.

orbeckst (approved):
This looks very good to me.

I am doing a small doc update and wait for CI but then it should be ready to be merged. 

Thank you very much for the contribution!

orbeckst (changes requested):
I reviewed after all the tests completed and coverage is still too low. 

It's good that you are trying to fail gracefully and defensive programming is good. But these failure cases need to be tested, too. I know that this is really tedious. But in order to keep the library maintainable in the future we need to decrease technical debt and untested code, especially with fairly complicated conditionals that require very specific knowledge, can be difficult to maintain. 

You can add broken test data to alchemlyb or you can generate the broken data as part of a test fixture. E.g., copy the real data set and then change or replace some of the original files in the copy, or write a file from an inline string with only a few lines. As long as you can trigger your exception, it's ok.

When you test exceptions, use
```python
with pytest.raises(ValueError, match="Error message"):
    do_stuff()
```
to test that the correct exception is raises _and_ that you see the error message that you're expecting.

jhenin:
@orbeckst I've merged updates from @ttjoseph - can you run the tests? Thanks!

ttjoseph:
@jhenin @orbeckst 

I have pushed another few commits to my branch that makes a test more specific and warns the user when lambda_idws must be inferred - please test that

orbeckst:
Hello @jhenin @ttjoseph , sorry that I hadn't started the tests sooner, notification slipped through.

I relaxed the permissions for running tests so that from now on only people new to GitHub require explicit permission to run tests, for everybody else they _should_ run automatically. I should have realized sooner that GitHub enabled this setting.

orbeckst:
@ttjoseph please check the failure for [Ubuntu Python 3.7](https://github.com/alchemistry/alchemlyb/pull/135/checks?check_run_id=3787538651). Not sure why it failed in 3.7 but passed in 3.8 and 3.9.

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -1,13 +1,25 @@
 """NAMD parser tests.
 
 """
```
orbeckst, line 3: Please do not use `tempfile` and manual cleaning up. Instead use the `tmpdir` and `tmp_path` fixtures of pytest https://docs.pytest.org/en/6.2.x/tmpdir.html.

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+        the output of running split() on the identified line and returns
+        a modified list of tokens that will be reassembled into the line to be
+        output.
+
+    Returns
+    -------
+    The name of a temporary file which the caller must unlink.
+    """
+
+    fh, fepout_out = mkstemp(suffix='.fepout.bz2', prefix='restarted_')
```
orbeckst, line 79: use `tmp_path`

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+                    if line.startswith(prefix):
+                        line = ' '.join(func(line.split())) + '\n'
+                f_out.write(line)
+    return fepout_out
+
+
+@pytest.fixture
+def restarted_dataset_inconsistent():
+    """Returns intentionally messed up dataset."""
+    # We explicitly call load_restarted() again because we are modifying the
```
orbeckst, line 94: You should be able to do this with normal function-scoped fixtures.

```python
@pytest.fixture
def restarted_dataset_inconsistent(restarted_dataset):
     filenames = restrarted_dataset['data']['both']
     ...
```
Here `restarted_dataset` will be a unique copy.

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+    temp_fnames = []
+    for i in range(len(filenames)):
+        fname = _corrupt_fepout(filenames[i], [('#Free', func_free_line)])
+        dataset['data']['both'][i] = fname
+        temp_fnames.append(fname)
+        # Only actually modify one window so we don't trigger the wrong exception
+        if changed is True:
+            break
+
+    yield dataset
```
orbeckst, line 115: just return

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+    for fname in temp_fnames:
```
orbeckst, line 116: no need to clean up when using `tmp_path` or `tmpdir`, pytest will clean up

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+        unlink(fname)
+
+
+@pytest.fixture
+def restarted_dataset_toomany_lambda2():
```
orbeckst, line 121: see above, use `restarted_dataset` fixture

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+        l[LAMBDA1_IDX_FREE] = '0.2'
+        return l
+
+    temp_fnames = []
+    for i in range(len(filenames)):
+        fname = _corrupt_fepout(filenames[i], [('#NEW', func_new_line), ('#Free', func_free_line)])
+        dataset['data']['both'][i] = fname
+        temp_fnames.append(fname)
+
+    yield dataset
```
orbeckst, line 144: see above, just return, and don't clean up (assuming everything uses `tmpdir`/`tmp_path`)

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+    for fname in temp_fnames:
+        unlink(fname)
+
+
+@pytest.fixture
+def restarted_dataset_toomany_lambda_idws():
```
orbeckst, line 150: see above, switch to using the pytest fixtures

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+            l[LAMBDA1_IDX_FREE], l[LAMBDA2_IDX_FREE] = '0.5', '0.6'
+        return l
+
+    temp_fnames = []
+    for i in range(len(filenames)):
+        fname = _corrupt_fepout(filenames[i], [('#NEW', func_new_line)])
+        dataset['data']['both'][i] = fname
+        temp_fnames.append(fname)
+
+    yield dataset
```
orbeckst, line 177: see above

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+    for fname in temp_fnames:
+        unlink(fname)
+
+
+@pytest.fixture
+def restarted_dataset_direction_changed():
```
orbeckst, line 183: use pytest tmp fixtures

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+    def func_free_line(l):
+        l[7], l[8] = l[8], l[7]
+        return l
+    
+    # Reverse the direction of lambdas for this window
+    idx_to_corrupt = filenames.index(sorted(filenames)[-3])
+    fname1 = _corrupt_fepout(filenames[idx_to_corrupt], [('#NEW', func_new_line), ('#Free', func_free_line)])
+    dataset['data']['both'][idx_to_corrupt] = fname1
+    yield dataset
+
```
orbeckst, line 201: see above

src/alchemlyb/tests/parsing/test_namd.py
```diff
@@ -36,13 +48,183 @@ def test_u_nk_idws():
...
+    unlink(fname1)
+
+
+def test_u_nk_restarted(restarted_dataset, restarted_dataset_direction_changed,
```
orbeckst, line 205: Make this one test for each of these different exceptions: one dataset, one exception test

orbeckst (changes requested):
Good work creating the faulty input files that trigger the exceptions. 

The main change that I'd like to see is to use pytest `tmpdir`/`tmp_path` instead of using your own `tmpfile`/clean up as this will (1) bring your code in line with the rest of the tests and (2) shorten your tests while also taking care of tests not influencing each other (by using function-scoped fixtures). Please also break the exception tests into separate tests. As a rule, one test function/method should test one distinct situation.

As I said in a comment, I am not sure why Python 3.7 failed while 3.8 & 3.9 passed. That's something to look into.

Overall, very good work making code and tests solid — you're almost there!!

ttjoseph:
@orbeckst I suspect the error on 3.7 is due to the randomly generated filenames for the corrupted .fepout files, so their filename sort order is not preserved and thus may be presented to the parser in arbitrary order, resulting in the wrong error. This depends on whatever the RNG seed is for mkstemp. So this could have randomly happened with any of the other Python versions. Now that I use tmp_path, I keep the old filenames and so the sort order should remain stable and not depend on RNG.

Anyway, this and other changes addressing your comments are in my repo, so for them to show up here I need @jhenin to merge them into his tree :)

orbeckst (approved):

orbeckst:
Thank you for addressing my last set of comments. This looking excellent now. I am just waiting for CI to be all green.

orbeckst:
@jhenin @ttjoseph  can you please also add your names to [AUTHORS](https://github.com/alchemistry/alchemlyb/blob/master/AUTHORS)?

I'll have look at docs and coverage.

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +108,162 @@ def extract_u_nk(fep_file, T):
...
+                    # Infer lambda_idws_at_start if it wasn't explictly included in this fepout.
+                    # If lambdas are in ascending order, choose the one before it
+                    # This can happen if user ran minimize before starting dynamics and didn't do firsttimestep 0,
+                    # because NAMD only emits the '#NEW' line on timestep 0 for some reason
+                    if has_idws and lambda_idws_at_start is None:
+                        l1_idx, l2_idx = all_lambdas.index(lambda1), all_lambdas.index(lambda2)
+                        if l1_idx > 0 and l1_idx < l2_idx: # Ascending lambdas
+                            lambda_idws_at_start = all_lambdas[l1_idx - 1]
+                        elif l2_idx < (len(all_lambdas) - 1) and l2_idx < l1_idx: # Descending lambdas
+                            lambda_idws_at_start = all_lambdas[l2_idx + 1]
```
orbeckst, line 203: When should this conditional be triggered? Is this a common or a rare thing?

At the moment this case is not covered by tests.

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +108,162 @@ def extract_u_nk(fep_file, T):
...
+                    win_de_arr = beta * np.asarray(win_de) # dE values
+                    win_ts_arr = np.asarray(win_ts) # timesteps
+
+                    # Infer lambda_idws_at_start if it wasn't explictly included in this fepout.
+                    # If lambdas are in ascending order, choose the one before it
+                    # This can happen if user ran minimize before starting dynamics and didn't do firsttimestep 0,
+                    # because NAMD only emits the '#NEW' line on timestep 0 for some reason
+                    if has_idws and lambda_idws_at_start is None:
+                        l1_idx, l2_idx = all_lambdas.index(lambda1), all_lambdas.index(lambda2)
+                        if l1_idx > 0 and l1_idx < l2_idx: # Ascending lambdas
```
orbeckst, line 200: This conditional has no `else` clause. Is that ok or should it fail if neither `if` nor `elif` is entered?

src/alchemlyb/parsing/namd.py
```diff
@@ -3,20 +3,103 @@
...
+            for line in f:
+                l = line.strip().split()
+                if l[0] not in ['#NEW', '#Free']:
+                    continue
+
+                # We might not have a #NEW line so make the best guess
+                if l[0] == '#NEW':
+                    lambda1, lambda2 = float(l[6]), float(l[8])
+                    lambda_idws = float(l[10]) if 'LAMBDA_IDWS' in l else None
+                elif l[0] == '#Free':
```
orbeckst, line 50: There is no `else` (and the case of neither `if` nor `elif` is never tested). Under which circumstances would `else` happen?

src/alchemlyb/parsing/namd.py
```diff
@@ -25,72 +108,162 @@ def extract_u_nk(fep_file, T):
...
+                    elif l[0] == 'FepE_back:':
+                        win_de_back.append(float(l[6]))
+                        win_ts_back.append(float(l[1]))
+
+                # Turn parsing on after line 'STARTING COLLECTION OF ENSEMBLE AVERAGE'
+                if '#STARTING' in l:
+                    parsing = True
+
+    if len(win_de) != 0 or len(win_de_back) != 0:
+        logger.warning('Trailing data without footer line (\"#Free energy...\"). Interrupted run?')
```
orbeckst, line 258: Is this harmless or should it error?

If it's harmless then add a `# pragma: nocover` to this block.

orbeckst (changes requested):
Some questions on code logic.

src/alchemlyb/parsing/namd.py
ttjoseph: This would occur when we parse a FEP calculation result where lambda progresses 1 -> 0 over the set of windows, rather than (to me) the usual case of 0 -> 1. This bit of code infers what the lambda_idws likely is, in the event no lambda_idws is present in this .fepout file - which can happen when a NAMD run is interrupted.

Hm, I guess we'd want to test a 1 -> 0 FEP calculation?

src/alchemlyb/parsing/namd.py
ttjoseph: This handles the special case where there are IDWS energies but no lambda_idws value in the current .fepout file, which can happen when NAMD is interrupted.

src/alchemlyb/parsing/namd.py
ttjoseph: Should be harmless, this parser is meant to handle interrupted runs, but the user should know what's going on.

src/alchemlyb/parsing/namd.py
orbeckst: If you as the expert say that this could happen in the wild then we want a test — yes, please.

src/alchemlyb/parsing/namd.py
orbeckst: Can you then just add your explanation as a comment to the code, please?

src/alchemlyb/parsing/namd.py
orbeckst: Ok, then mark it as not covered by tests. I *think* you can just modify the line
```python
    if len(win_de) != 0 or len(win_de_back) != 0:   # pragma: no cover
```
and it should exclude the whole block. But see https://coverage.readthedocs.io/en/latest/excluding.html for details.

src/alchemlyb/parsing/namd.py
ttjoseph: This function only extracts the lambda values from a set of .fepout files. The lambda values are only in a small subset of lines in each file, so the `else` case is a `FepEnergy` or `FepE_back` line, which also contains an energy value for the timestep, which we don't care about in this function.

orbeckst: Can you please also add an explanatory comment? That would be sufficient. Thanks.

src/alchemlyb/parsing/namd.py
ttjoseph: OK. In this case the easiest thing for me is to add another test set in alchemtest. The format of NAMD .fepout files that use IDWS makes it difficult to modify the existing test set on the fly as is done to trigger the other failure cases: lambda values are in both header and footer, but lambda_idws is not in footer. So swapping them around across fragments of .fepout files that would arise from an interrupted and restarted simulation would make the test code more complex than the code it's meant to test.

orbeckst: If you add it to alchemtest and ping me I'll review it as quickly as possible and try to get it in ASAP.

ttjoseph:
This seems to be another bug in the test related to filesystem file ordering. On my machine they happen to be read in lexicographic order by filename but not on GitHub CI. Another commit pushed...thanks in advance, @jhenin

ttjoseph:
I've added one (last?) pragma no cover as a PR to @jhenin's tree

orbeckst:
@ttjoseph and @jhenin , congratulations 👏 👏 👏 , your IDWS NAMD parser is in alchemlyb together with all the robustness improvements. 

Thank you for not giving up and seeing it through — through >100 comments and numerous "almost there" reviews. Much appreciated!!!

PR #141, opened 2021-06-10
adopt NEP29

### LINKED ISSUE #140: adopt NEP29 for support of Python versions
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) describes a time-based deprecation policy for scientific python projects for which versions of Python (and numpy) to support. 

> When a project releases a new major or minor version, we recommend that they support at least all minor versions of Python introduced and released in the prior 42 months from the anticipated release date with a minimum of 2 minor versions of Python, and all minor versions of NumPy released in the prior 24 months from the anticipated release date with a minimum of 3 minor versions of NumPy.

We also depend on pandas but the [panda's Python Policy](https://pandas.pydata.org/docs/development/policies.html#python-support) is not clear, except
> pandas will only drop support for specific Python versions (e.g. 3.6.x, 3.7.x) in pandas major or minor releases.

For example, pandas 1.2 does not  support Python 3.6 anymore (which reaches [end of support](https://www.python.org/downloads/)  2021-12-23) and [fell out of the NEP29 range in Nov 2020](https://numpy.org/neps/nep-0029-deprecation_policy.html#detailed-description).

**alchemlyb heavily relies on pandas and numpy** (and **pymbar**). To keep the maintenance burden manageable we propose to effectively follow the support roadmap of our major upstream dependencies. 

----

For the new units code #125 we really want to use pandas ≥ 1.2 because earlier support for pandas.DataFrame.attrs is very buggy. This means that we would drop Python 3.6 support for the next minor release (0.5.0) in addition to having already dropped 2.7 and 3.5 #130.  

----
Maybe useful [nep29-calculator](https://pypi.org/project/nep29/)

orbeckst:
- close #140
- drop Python 3.6
- require pandas >= 1.2
- update setp.py metadata
- update CHANGES
- updated docs (add note to Development Model)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -39,8 +39,17 @@ Development is done in the open on `GitHub`_.
 
 Software engineering best-practices are used throughout, including continuous integration testing via Travis CI, up-to-date documentation, and regular releases.
 
-.. _`GitHub`: https://github.com/alchemistry/alchemlyb
+.. Note::
+   With release 0.5.0, the alchemlyb project adopted `NEP 29`_ to determine which versions of Python and NumPy_ will be supported.
+   When we release a new major or minor version, alchemlyb will support *at least all minor versions of Python introduced and released in the prior 42 months from the release date with a minimum of 2 minor versions of Python*, and *all minor versions of NumPy released in the prior 24 months from the anticipated release date with a minimum of 3 minor versions of NumPy*.
 
+The pandas_ package (one of our other primary dependencies) also follows `NEP 29`_ so this support policy will make it easier to maintain **alchemlyb** in the future.
+
+   
+.. _`GitHub`: https://github.com/alchemistry/alchemlyb
+.. _`NEP 29`: https://numpy.org/neps/nep-0029-deprecation_policy.html
+.. _NumPy: https://numpy.org
+.. _pandas: https://pandas.pydata.org/
 
 .. _contact:
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
     - MPLBACKEND=agg
 
 python:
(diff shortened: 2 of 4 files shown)


docs/index.rst
```diff
@@ -39,8 +39,17 @@ Development is done in the open on `GitHub`_.
 
 Software engineering best-practices are used throughout, including continuous integration testing via Travis CI, up-to-date documentation, and regular releases.
 
-.. _`GitHub`: https://github.com/alchemistry/alchemlyb
+.. Note::
+   With release 0.5.0, the alchemlyb project adopted `NEP 29`_ to determine which versions of Python and NumPy_ will be supported.
+   When we release a new major or minor version, alchemlyb will support *at least all minor versions of Python introduced and released in the prior 42 months from the release date with a minimum of 2 minor versions of Python*, and *all minor versions of NumPy released in the prior 24 months from the anticipated release date with a minimum of 3 minor versions of NumPy*.
+
+The pandas_ package (one of our other primary dependencies) also follows `NEP 29`_ so this support policy will make it easier to maintain **alchemlyb** in the future.
```
orbeckst, line 46: @xiki-tempula do you have a link that indicates that pandas adopted NEP29?

xiki-tempula: @orbeckst It seems that they just dropped it https://github.com/pandas-dev/pandas/pull/35214

orbeckst: Thanks, it's not an officially written "We adopt NEP29" (otherwise they would have done so in their policy) so I am not going to link it but it's as close as we get insight into their intentions.

xiki-tempula (approved):
The change looks good to me.

PR #142, opened 2021-06-15
Added initial GHA CI

dotsdl:
Executes tests on `ubuntu-latest` with Python versions:
- 3.7
- 3.8
- 3.9


dotsdl:
@orbeckst is there anything missing in your view?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - name: Install alchemtest
+      shell: bash -l {0}
+      run: |
+        python -m pip install --only-binary=numpy numpy # Otherwise this would take ages
+        python -m pip install https://github.com/alchemistry/alchemtest/archive/master.zip
+
+    - name: Install package
+      shell: bash -l {0}
+      run: |
+        python -m pip install -e .
```
orbeckst, line 56: Do we really want a developer installation or is better to do a standard installation as this also checks that everything is copied appropriately? I don't remember the rationale for the developer installation, though.

orbeckst (approved):
Thank you! Looks great, just one optional question about developer installation (see below).

Also, why is Travis still running on this PR??

(We'll open more issues for testing doc building and package building.)

orbeckst:
In my review I forgot the badge in the README. Please update — I am not going to change my review to blocking for that.

In any case it looks as if I have to merge with admin power because Travis CI is blocking this merge — possibly does not want to be replaced ;-).

.github/workflows/ci.yaml
dotsdl: The rationale is that you're testing the same `alchemlyb` as the source files, but in this case doesn't matter since it's ephemeral anyway. I'll remove the `-e`.

orbeckst:
I removed "Travis CI required" from branch protection. That was the issue with Travis.

dotsdl:
I also removed the Travis CI webhook. Did the same for codecov since we're pushing from each of these builds.

dotsdl:
Although on second thought we might want to keep codecov for its diffs, but how does it get them if not from running the tests?

orbeckst:
I _think_ codecov only gets it through the tests.

orbeckst:
I've got to be elsewhere. Please feel free to merge the PR yourself when you deem it done.

Thank you so much, @dotsdl !

dotsdl:
Okay! Just got the badge added to the README. If these pass, I'll merge. We can fix anything else missing in post. :wink:

PR #146, opened 2021-07-04
Update estimators-ti.rst

argentumgun:
there exists an incorrect code: alchemlyb.concat should be repalced by pd.concat


orbeckst:
@argentumgun the use of `alchemlyb.concat` is  necessary since issue #125 so that metadata are correctly transferred with dataframes. @xiki-tempula can say more.

xiki-tempula:
@argentumgun Thanks for the PR. I wonder what issue is this PR trying to fix? 
Before https://github.com/pandas-dev/pandas/pull/42252 is merged and the new pandas is released, the current pandas cannot propagate the metadata and alchemlyb.concat is created to patch this problem.

argentumgun:
sorry，maybe the version I installed needs to be upgraded

PR #148, opened 2021-07-09
Fix pandas 1.3/1.4 support

### LINKED ISSUE #147: tests for Python 3.7 started failing with 'TypeError: Argument 'obj' has incorrect type (expected list, got FrozenList)'
The scheduled tests for Python 3.7 are failing now, see https://github.com/alchemistry/alchemlyb/runs/2983317549?check_suite_focus=true

Lots of errors like to following:

```
 ______________________________ Test_Units.test_ti ______________________________

self = <alchemlyb.tests.test_ti_estimators.Test_Units object at 0x7f8ae3af9050>
dhdl =                           fep
time    fep-lambda           
0.0     0.0         13.390058
10.0    0.0          9.23137...338
39980.0 1.0         -1.184511
39990.0 1.0         -1.273484
40000.0 1.0         -0.416037

[20005 rows x 1 columns]

    def test_ti(self, dhdl):
>       ti = TI().fit(dhdl)

src/alchemlyb/tests/test_ti_estimators.py:177: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/alchemlyb/estimators/ti_.py:64: in fit
    dl = means.reset_index()[means.index.names[:]].diff().iloc[1:].values
/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/pandas/core/frame.py:3445: in __getitem__
    if com.is_bool_indexer(key):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

key = FrozenList(['fep-lambda'])

    def is_bool_indexer(key: Any) -> bool:
        """
        Check whether `key` is a valid boolean indexer.
    
        Parameters
        ----------
        key : Any
            Only list-likes may be considered boolean indexers.
            All other types are not considered a boolean indexer.
            For array-like input, boolean ndarrays or ExtensionArrays
            with ``_is_boolean`` set are considered boolean indexers.
    
        Returns
        -------
        bool
            Whether `key` is a valid boolean indexer.
    
        Raises
        ------
        ValueError
            When the array is an object-dtype ndarray or ExtensionArray
            and contains missing values.
    
        See Also
        --------
        check_array_indexer : Check that `key` is a valid array to index,
            and convert to an ndarray.
        """
        if isinstance(key, (ABCSeries, np.ndarray, ABCIndex)) or (
            is_array_like(key) and is_extension_array_dtype(key.dtype)
        ):
            if key.dtype == np.object_:
                key = np.asarray(key)
    
                if not lib.is_bool_array(key):
                    na_msg = "Cannot mask with non-boolean array containing NA / NaN values"
                    if lib.infer_dtype(key) == "boolean" and isna(key).any():
                        # Don't raise on e.g. ["A", "B", np.nan], see
                        #  test_loc_getitem_list_of_labels_categoricalindex_with_na
                        raise ValueError(na_msg)
                    return False
                return True
            elif is_bool_dtype(key.dtype):
                return True
        elif isinstance(key, list):
            # check if np.array(key).dtype would be bool
>           return len(key) > 0 and lib.is_bool_list(key)
E           TypeError: Argument 'obj' has incorrect type (expected list, got FrozenList)
```

I don't know what the root problem is but perhaps something changed in pandas and we're picking up the latest version?

xiki-tempula:
Fix #147 

Due to https://github.com/pandas-dev/pandas/issues/42433, this will need some fix from upstream.
I have done a patch to fix the problem.

### DIFF
--- a/src/alchemlyb/estimators/ti_.py
+++ b/src/alchemlyb/estimators/ti_.py
@@ -1,5 +1,3 @@
-import pickle
-
 import numpy as np
 import pandas as pd
 
--- a/src/alchemlyb/estimators/ti_.py
+++ b/src/alchemlyb/estimators/ti_.py
@@ -1,3 +1,5 @@
+import pickle
+
 import numpy as np
 import pandas as pd
 
@@ -61,7 +63,7 @@ class TI(BaseEstimator):
         l_types = dHdl.index.names[1:]
 
         # obtain vector of delta lambdas between each state
-        dl = means.reset_index()[means.index.names[:]].diff().iloc[1:].values
+        dl = means.reset_index()[list(means.index.names[:])].diff().iloc[1:].values
 
         # apply trapezoid rule to obtain DF between each adjacent state
         deltas = (dl * (means.iloc[:-1].values + means.iloc[1:].values)/2).sum(axis=1)
@@ -133,7 +135,7 @@ class TI(BaseEstimator):
         # get the lambda names
         l_types = self.dhdl.index.names
         # obtain bool of changed lambdas between each state
-        lambdas = self.dhdl.reset_index()[l_types]
+        lambdas = self.dhdl.reset_index()[list(l_types)]
         diff = lambdas.diff().to_numpy(dtype='bool')
         # diff will give the first row as NaN so need to fix that
         diff[0, :] = diff[1, :]
--- a/src/alchemlyb/estimators/ti_.py
+++ b/src/alchemlyb/estimators/ti_.py
@@ -61,6 +61,8 @@ class TI(BaseEstimator):
         l_types = dHdl.index.names[1:]
 
         # obtain vector of delta lambdas between each state
+        # Fix issue #148, where for pandas != 1.3.0
+        # dl = means.reset_index()[means.index.names[:]].diff().iloc[1:].values
         dl = means.reset_index()[list(means.index.names[:])].diff().iloc[1:].values
 
         # apply trapezoid rule to obtain DF between each adjacent state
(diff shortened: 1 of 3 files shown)


src/alchemlyb/estimators/ti_.py
```diff
@@ -61,7 +61,7 @@ def fit(self, dHdl):
         l_types = dHdl.index.names[1:]
 
         # obtain vector of delta lambdas between each state
-        dl = means.reset_index()[means.index.names[:]].diff().iloc[1:].values
+        dl = means.reset_index()[list(means.index.names[:])].diff().iloc[1:].values
```
orbeckst, line 64: add a comment that we undo this fix once pandas is fixed

src/alchemlyb/estimators/ti_.py
```diff
@@ -133,7 +133,7 @@ def separate_dhdl(self):
         # get the lambda names
         l_types = self.dhdl.index.names
         # obtain bool of changed lambdas between each state
-        lambdas = self.dhdl.reset_index()[l_types]
+        lambdas = self.dhdl.reset_index()[list(l_types)]
```
orbeckst, line 136: add a comment that we undo this fix once pandas is fixed

orbeckst (approved):
Good for a quick fix. Eventually we want to revert the fix and just _exclude_ buggy versions of pandas in setup.py. 

I see that pandas has already a bug fix PR https://github.com/pandas-dev/pandas/pull/42469 and the next patch release will likely have it fixed.

Please add comments to the hot fix lines and also add a comment to setup.py for which versions of pandas we'll eventually exclude. Once that's done, merge, please. Thank you!

orbeckst:
Note: the fix PR https://github.com/pandas-dev/pandas/pull/42469 is in their 1.3.1 milestone so we will just have to exclude 1.3.0 in the [install_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#install-requires) list https://github.com/alchemistry/alchemlyb/blob/dce28930bc513bc7cde50d3597b7cdb2d26ee1ad/setup.py#L46 as
```python
"pandas>=1.2,!=1.3.0"
```

PR #162, opened 2021-08-30
Make it that the MBAR will take in maximum_iterations argument.

xiki-tempula:
Currently, at least `maximum_iterations` is not really being used during the MBAR solver.
So I changed the MBAR to make sure that it is taken into account as is suggested in https://github.com/choderalab/pymbar/issues/419#issuecomment-803714103.

Note the doc says other things though https://pymbar.readthedocs.io/en/master/mbar.html.

I cannot write a test to test this as `pymbar.mbar.MBAR` doesn't have a public attribute that echoes this assignment.

It is hard to raise an issue either as the bug will show itself when MBAR didn't converge within the default 250 iterations and most of the calculations converge in less than 250 iterations.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,7 @@ Fixes
     compatible with pandas.concat() (issue #150, PR #152).
   - Fix the support for pandas >= 1.3 by skipping 1.3.0 (issue #147, PR #148).
   - Fix separate_dhdl not work for multiple columns (issue #149, PR #151).
+  - MBAR estimator now correctly passes max_iterations to pymbar.MBAR. (PR #162)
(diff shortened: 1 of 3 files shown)


orbeckst (changes requested):
Can you just add an entry to CHANGES as a Fixed: "MBAR estimator now correctly passes max_iterations to pymbar.MBAR" ?

Did you test that it works with a local example that failed previously? If you can confirm that you validated locally, I am ok with the PR, even without tests.

xiki-tempula:
Thanks for the review. It works with the local example, I will merge this PR.

PR #163, opened 2021-09-11
Implement the decorrelator

xiki-tempula:
As is discussed in https://github.com/alchemistry/alchemlyb/discussions/159. I have implemented the decorrelators using the logic of the alchemical analysis.


src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -225,3 +226,32 @@ def test_equilibrium_detection(self, dhdl):
...
+def test_decorrelate_dhdl_single_l():
+    dataset = load_benzene()
+    dhdl = extract_dHdl(dataset['data']['Coulomb'][0], 310)
+    assert len(decorrelate_dhdl(dhdl, drop_duplicates=True,
+                                sort=True)) == 2001
+
+def test_decorrelate_dhdl_multiple_l():
+    dataset = load_ABFE()
+    dhdl = extract_dHdl(dataset['data']['complex'][0], 310)
+    assert len(decorrelate_dhdl(dhdl,)) == 501
```
orbeckst, line 257: new line?

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -225,3 +226,32 @@ def test_equilibrium_detection(self, dhdl):
...
+@pytest.mark.parametrize(('method', 'size'), [('dhdl', 501),
+                                              ('dhdl_all', 1001),
+                                              ('dE', 334)])
+def test_decorrelate_u_nk_multiple_l(method, size):
+    dataset = load_ABFE()
+    u_nk = extract_u_nk(dataset['data']['complex'][-1], 310)
+    assert len(decorrelate_u_nk(u_nk, method=method,)) == size
+
+def test_decorrelate_dhdl_single_l():
+    dataset = load_benzene()
```
orbeckst, line 249: fixture

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -225,3 +226,32 @@ def test_equilibrium_detection(self, dhdl):
...
+    dhdl = extract_dHdl(dataset['data']['Coulomb'][0], 310)
+    assert len(decorrelate_dhdl(dhdl, drop_duplicates=True,
+                                sort=True)) == 2001
+
+def test_decorrelate_dhdl_multiple_l():
+    dataset = load_ABFE()
```
orbeckst, line 255: fixture

orbeckst (changes requested):
Can you please explain what the three methods are? Neither the user nor I really knows what the code is supposed to be doing. Once I know what's supposedt o happen, I can review the code in a more meaningful manner. 

The other comments are all pretty minor.

Please start _a new section in CHANGES_ for 0.6.0 and add an entry.

Thanks for adding the functionality!

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -7,6 +7,104 @@
...
+def decorrelate_dhdl(df, **kwargs):
+    """Subsample a dhdl DataFrame.
+
+    This is wrapper function around the
+    :func:`~alchemlyb.preprocessing.subsampling.statistical_inefficiency`.
+
+    Parameters
+    ----------
+    df : DataFrame
+        DataFrame to subsample according to the selected method.
```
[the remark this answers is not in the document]
xiki-tempula: There is only one way (I think) to decorrelate the dhdl so I haven't put the method keyword in.

xiki-tempula:
@orbeckst Thank you very much for the help. I'm struggling with the pytest.fixture.

Normally, I use fixture or mark.parametrize but I realised now that there might be a problem with using both of them

@pytest.fixture()
def data():
    return [1, 2]

@pytest.mark.parametrize(('data', 'value'), [(data[0], 1), (data[1], 2)])
def test_value(data, value):
    assert data == value

will not work, where it will complain that the pytest fixture is a function and can not be subscripted.

src/alchemlyb/preprocessing/subsampling.py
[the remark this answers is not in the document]
orbeckst: of course, my fault

orbeckst:
For the fixture, I meant something along the following lines
```python
@pytest.fixture
def dataset():
   return data.load_ABFE()

@pytest,makr.parametrize("value,ref", ...)
def test_something(dataset, value, ref):
   ...
```
i.e., just not calling data.load_* inside but providing the data as a fixture. I'd _think_ that this won't collide.

xiki-tempula:
@orbeckst Thanks for the advice.
There is an existing test 
https://github.com/alchemistry/alchemlyb/blob/093ad87eca736f724c9a4b4d6843043ea9861f69/src/alchemlyb/tests/test_preprocessing.py#L191
which uses gmx_benzene_dHdl as a function, and I would need gmx_benzene_dHdl to be a fixture for my test. 
I wonder what is your advice for transforming this test https://github.com/alchemistry/alchemlyb/blob/093ad87eca736f724c9a4b4d6843043ea9861f69/src/alchemlyb/tests/test_preprocessing.py#L192? Thank you

orbeckst:
Perhaps something like
```python
@pytest.fixture(scope="class")   # class scope only if the data are never modified
def dataset(self):
   return gmx_benzene_dHdl()

@pytest.mark.paramtrize("stop,step", [(20, None), (None, -1)])
def test_series(self, dataset, start, stop):
   series = dataset[:stop:step]
   with pytest.raises(ValueError):
         self.slicer(dataset, series=series)
```

The advantage would be that with a class-scoped fixture, `gmx_benzene_dHdl()` is only loaded once.

xiki-tempula:
@orbeckst Thank you very much for the advice, there is another existing test that I'm not sure of how to make it into a fixture.

https://github.com/alchemistry/alchemlyb/blob/093ad87eca736f724c9a4b4d6843043ea9861f69/src/alchemlyb/tests/test_preprocessing.py#L50

```
    @pytest.mark.parametrize(('data', 'size'), [(gmx_benzene_dHdl, 661),
                                                (gmx_benzene_u_nk, 661)])
    def test_basic_slicing(self, data, size):
        assert len(self.slicer(data(), lower=1000, upper=34000, step=5)) == size
```

orbeckst:
I think I would just leave this one. 

There's a way to parametrize fixtures but it's not cleaner than what you already have for this test. If we don't re-use a fixture and it's not going to be easier to read then leave it, I'd say.

DRACO DORMIENS NUMQUAM TITILLANDUS

xiki-tempula:
@orbeckst Thanks for the advice. I have eventually left the old test untouched and have only applied the fixture to the new tests.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -7,6 +7,121 @@
...
+
+    Returns
+    -------
+    DataFrame
+        `df` subsampled according to selected `method`.
+
+    Note
+    ----
+    The default of ``True`` for  `drop_duplicates` and `sort` should result in robust decorrelation
+    but can loose data.
```
orbeckst, line 44:
```suggestion
    but can loose data.

    .. versionadded:: 0.6.0
```

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -7,6 +7,121 @@
...
+def decorrelate_u_nk(df, method='dhdl', drop_duplicates=True,
+                     sort=True, **kwargs):
+    """Subsample a u_nk DataFrame based on the selected method.
+
+    The method can be either 'dhdl_all' (obtained as a sum over all energy
+    components) or 'dhdl' (obtained as the energy components that are
+    changing; default) or 'dE'. In the latter case the energy differences
+    dE_{i,i+1} (dE_{i,i-1} for the last lambda) are used.
+
+    This is wrapper function around the
```
orbeckst, line 19:
```suggestion
    This is a wrapper function around the function
```

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -7,6 +7,121 @@
...
+            series = df.iloc[:, index - 1]
+    else: # pragma: no cover
+        raise ValueError(
+            'Decorrelation method {} not found.'.format(method))
+    return statistical_inefficiency(df, series, **kwargs)
+
+def decorrelate_dhdl(df, drop_duplicates=True, sort=True, **kwargs):
+    """Subsample a dhdl DataFrame.
+
+    This is wrapper function around the
```
orbeckst, line 93:
```suggestion
    This is a wrapper function around the function
```

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -7,6 +7,121 @@
...
+    Returns
+    -------
+    DataFrame
+        `df` subsampled.
+
+    Note
+    ----
+    The default of ``True`` for  `drop_duplicates` and `sort` should result in robust decorrelation
+    but can loose data.
+    """
```
orbeckst, line 117:
```suggestion


    .. versionadded:: 0.6.0
    """
```

orbeckst (approved):
lgtm, will just add my minor fixes

PR #172, opened 2021-10-23
implement auto mbar 

### LINKED ISSUE #170: About a more robust MBAR solver
Due to https://github.com/choderalab/pymbar/issues/419, a number of downstream packages such as https://github.com/luancarvalhomartins/PyAutoFEP have to stick with alchemlyb 0.3.0 & pymbar 3.0.3. I'm thinking if it would be nice to have a more robust MBAR solver so that the downstream packages could use the latest version.

The pseudocode might be that 

```
class MBAR(method=None):
    def fit(self,):
        if method is not None:
            pymbar.MBAR(method=method)
        else:
            try:
                pymbar.MBAR(method='adaptive')
            except:
                pymbar.MBAR(method='BFGS')
```

I'm aware that pymbar is working on solving this problem. I think it won't hurt to have an extra layer of protection at alchemlyb level.

xiki-tempula:
Fix #170

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,7 +18,6 @@ The rules for this file:
   * 0.6.0
 
 Changes
-  - Automatic MBAR solver with fall back (issue #170, PR #172)
 
 Enhancements
   - support for Interleaved Double-Wide Sampling (IDWS) in NAMD (PR #135).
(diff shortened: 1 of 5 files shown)


xiki-tempula:
@orbeckst I think the new logging look fine. I was under the impression that for newer python (>py2.5) string formating is encouraged compared with %. I wonder what is your thought?

orbeckst:
logging supports % strings where the values are arguments to the logger, see https://docs.python.org/3/howto/logging.html#logging-variable-data for 3.10 — this is how it's done for loggers

PR #173, opened 2021-10-28
fixed AMBER parser logging

orbeckst:
- log to module logger and not the root logger
- removed redundant ERROR and WARNING text
- removed redundant trailing newlines

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" https://semver.org
 
 ------------------------------------------------------------------------------
-??/??/2021 schlaicha, xiki-tempula, jhenin, ttjoseph
+??/??/2021 schlaicha, xiki-tempula, jhenin, ttjoseph, orbeckst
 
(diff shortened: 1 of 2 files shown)


xiki-tempula (approved):
LGTM. 
A bit confused when I see that
`logger = logging.getLogger("alchemlyb.parsers.Amber")`
was added before this commit.

xiki-tempula:
@orbeckst Shall I just squash and merge this commit or do you prefer to do the squash and merge it?

orbeckst:
What do you mean, what's confusing?

The loggers had been in the original code but they did not use the specific logger but the root logger.

If you think anything needs changing, just ask for revisions ;-).

orbeckst:
You squash-merge when you're satisfied. I just assigned you.

xiki-tempula:
I mean when I saw that logger is used instead of logging, I would assume that one would need to create a logger, but the creation of the logger is not in this PR so I was initially quite confused. Then I noticed that the logger was created but never used.

PR #183, opened 2021-12-30
Implement a more robust gmx parser

### LINKED ISSUE #126: Remove corrupted lines in xvg files
Hi,

when I want to analyze free energy differences while simulations are still running, the last line of the xvg files is often corrupted (not fully written yet) and alchemlyb fails to do the analysis. Alchemical analysis has a feature that repairs those files, so I usually run that first and then run alchemlyb on the repaired xvg files. Is it possible to move that feature into alchemlyb?
https://github.com/MobleyLab/alchemical-analysis/blob/master/alchemical_analysis/utils/corruptxvg.py

### LINKED ISSUE #171: Gromacs parser cannot handle incomplete lines with -
For an incomplete line like

`1 -100 -10 -`

The last `-` will be giving problems as it cannot be cast as a float during 
https://github.com/xiki-tempula/alchemlyb/blob/093ad87eca736f724c9a4b4d6843043ea9861f69/src/alchemlyb/parsing/gmx.py#L308

where dtype=np.float64 would enforce it to be a float and `-` will give

```
File "/Users/XXX/opt/miniconda3/envs/mda_dev/lib/python3.7/site-packages/pandas/io/parsers.py", line 610, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "/Users/XXX/opt/miniconda3/envs/mda_dev/lib/python3.7/site-packages/pandas/io/parsers.py", line 468, in _read
    return parser.read(nrows)
  File "/Users/XXX/opt/miniconda3/envs/mda_dev/lib/python3.7/site-packages/pandas/io/parsers.py", line 1057, in read
    index, columns, col_dict = self._engine.read(nrows)
  File "/Users/XXX/opt/miniconda3/envs/mda_dev/lib/python3.7/site-packages/pandas/io/parsers.py", line 2061, in read
    data = self._reader.read(nrows)
  File "pandas/_libs/parsers.pyx", line 756, in pandas._libs.parsers.TextReader.read
  File "pandas/_libs/parsers.pyx", line 771, in pandas._libs.parsers.TextReader._read_low_memory
  File "pandas/_libs/parsers.pyx", line 850, in pandas._libs.parsers.TextReader._read_rows
  File "pandas/_libs/parsers.pyx", line 982, in pandas._libs.parsers.TextReader._convert_column_data
  File "pandas/_libs/parsers.pyx", line 1056, in pandas._libs.parsers.TextReader._convert_tokens
ValueError: could not convert string to float: '-'
```

xiki-tempula:
Fix #171 
Fix #126

I tried to do this as a preprocessing step but I cannot get it to fix the issues that I tested in the test files, so I implement it inside the parser.


src/alchemlyb/parsing/gmx.py
```diff
@@ -20,6 +20,8 @@ def extract_u_nk(xvg, T):
         Path to XVG file to extract data from.
     T : float
         Temperature in Kelvin the simulations sampled.
+    filter : bool
+        Filter out the lines that cannot be parsed.
```
orbeckst, line 24: described the types of lines that are filtered

src/alchemlyb/parsing/gmx.py
```diff
@@ -289,6 +293,8 @@ def _extract_dataframe(xvg, headers=None):
     headers: dict
        headers dictionary to search header information, reduced I/O by
        reusing if it is already parsed. Direct access by key name
+    filter : bool
+        Filter out the lines that cannot be parsed.
```
orbeckst, line 297: describe types of lines that are filtered

src/alchemlyb/parsing/gmx.py
orbeckst: also mention if `filter=True` is slower; add any information that users need to have to make a decision if to use `filter=True`

src/alchemlyb/parsing/gmx.py
```diff
@@ -304,9 +310,22 @@ def _extract_dataframe(xvg, headers=None):
...
+                memory_map=True, on_bad_lines='skip')
+        # If names=cols is passed to read_csv, rows with more than the
+        # designated columns will be truncated and used instead of discarded.
+        df.rename(columns={i: name for i, name in enumerate(cols)}, inplace=True)
+        # If dtype=np.float64 and float_precision='high' are passed to read_csv,
+        # 12.345.56 and - cannot be read.
+        df = df.apply(pd.to_numeric, errors='coerce')
+        # drop duplicate
+        df.dropna(inplace=True)
+    else:
```
orbeckst, line 324: logically, I'd do
```python
if not filter:
  # default, fast, assumes clean files
  ...
else:
  # slower ... etc etc more comments ...
```

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+    '''Test dropping the row that is wrong in different way'''
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def data():
+        dhdl = extract_dHdl(load_benzene()['data']['Coulomb'][0], 310)
+        with bz2.open(load_benzene()['data']['Coulomb'][0], "rt") as bz_file:
+            text = bz_file.read()
+        return text, len(dhdl)
+
+    def test_sanity(self, data, tmpdir):
```
orbeckst, line 212: Instead of `tmpdir`, it's now [recommended to use `tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html) and then 
```python
new_text = tmp_path / "text.xvg"
```

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        '''Test if the test routine is working.'''
+        text, length = data
+        new_text = tmpdir.join('text.xvg')
+        new_text.write(text)
+        dhdl = extract_dHdl(tmpdir.join('text.xvg'), 310)
```
orbeckst, line 217: use variable name
```suggestion
        dhdl = extract_dHdl(new_text, 310)
```

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        assert len(dhdl) == length
+
+    def test_truncated_row(self, data, tmpdir):
```
orbeckst, line 220: tmpdir --> tmp_path (optional... we'll have to do it eventually)

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        '''Test the case where the last row has been truncated.'''
+        text, length = data
+        new_text = tmpdir.join('text.xvg')
+        new_text.write(text + '40010.0 27.0\n')
+        dhdl = extract_dHdl(tmpdir.join('text.xvg'), 310, filter=True)
+        assert len(dhdl) == length
+
+    def test_truncated_number(self, data, tmpdir):
```
orbeckst, line 228: tmp_path

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        '''Test the case where the last row has been truncated and a - has
+        been left.'''
+        text, length = data
+        new_text = tmpdir.join('text.xvg')
+        new_text.write(text + '40010.0 27.0 -\n')
+        dhdl = extract_dHdl(tmpdir.join('text.xvg'), 310, filter=True)
```
orbeckst, line 234: use variable name

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        assert len(dhdl) == length
+
+    def test_weirdnumber(self, data, tmpdir):
```
orbeckst, line 237: tmp_path

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        '''Test the case where the last number has been appended a weird
+        number.'''
+        text, length = data
+        new_text = tmpdir.join('text.xvg')
+        # Note the 27.040010.0 which is the sum of 27.0 and 40010.0
+        new_text.write(text + '40010.0 27.040010.0 27.0 0.0 6.7 13.5 20.2 27.0 0.7 27.0 0.0 6.7 '
+                       '13.5 20.2 27.0 0.7\n')
+        dhdl = extract_dHdl(tmpdir.join('text.xvg'), 310, filter=True)
```
orbeckst, line 245: variable name instead of `tmpdir.join('text.xvg')`

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        assert len(dhdl) == length
+
+    def test_too_many_cols(self, data, tmpdir):
```
orbeckst, line 248: tmp_path

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -195,3 +198,60 @@ def test_extract_dHdl_unit():
...
+        '''Test the case where the row has too many columns.'''
+        text, length = data
+        new_text = tmpdir.join('text.xvg')
+        new_text.write(text +
+                       '40010.0 27.0 0.0 6.7 13.5 20.2 27.0 0.7 27.0 0.0 6.7 13.5 20.2 27.0 0.7\n')
+        dhdl = extract_dHdl(tmpdir.join('text.xvg'), 310, filter=True)
```
orbeckst, line 254: variable name

orbeckst (changes requested):
Looking very good. 

1. Primarily we need more documentation. 
2. add an entry to CHANGES (bump current 0.6.1 to 0.7.0) as Enhancement (you are adding a new kwarg)

Minor suggestions for the tests.

xiki-tempula:
@orbeckst Thanks for the review. I made the changes. I wonder how should I bump the version? sorry, I'm not too familiar with the versioner.

orbeckst:
Just change the entry in the CHANGELOG.

The *actual* version change happens on release anyway.

CHANGES
```diff
@@ -13,7 +13,15 @@ The rules for this file:
...
+
+*/*/2022 xiki-tempula
+
+  * 0.7.0
+
+Enhancements
+  - Add filter function to gmx.extract to make it more robust.
+
+03/04/2022 xiki-tempula
   * 0.6.1
```
orbeckst, line 25: remove because we won't release a 0.6.1, we go straight to 0.7.0
```suggestion
```

src/alchemlyb/parsing/gmx.py
orbeckst: Is it slower, and if so, by how much?

orbeckst (changes requested):
As a user I would currently always set `filter=True` because there's no compelling reason to use `False`. There should be some docs that make the user understand why the default is `False`. I assume that this comes down to speed but before saying this, we need to have an idea how much the filtering slows it down. Can you do a quick check for some reasonable input and report the timing? Then update the docs.

xiki-tempula:
@orbeckst Thanks for the review. I have changed the default to True. And add versionchange
It seems that filtering is actually faster but this might just be random fluctuation.

```
dataset = alchemtest.gmx.load_benzene()
time_list = []
for i in range(100):
    start = time.time()
    alchemlyb.concat([gmx.extract_u_nk(filename, T=300, filter=True) for filename in dataset['data']['Coulomb']])
    time_list.append(time.time()-start)
print(np.mean(time_list))
```
gives 0.14853424549102784
Changing the filter=False gives 0.152081116437912

orbeckst (approved):
Thanks for addressing my comments and checking performance.

I am updating the CHANGELOG after submitting this review to make clear that we are changing default behavior.

orbeckst:
thanks @xiki-tempula !

PR #184, opened 2021-12-30
Error in the load_expanded_ensemble_case_1

xiki-tempula:
Note the dataframes in load_expanded_ensemble_case_1 has NaNs in it as dropping the na will create a problem in the tests.


xiki-tempula:
fail-fast is set to false to allow all tests to run.

orbeckst:
In principle using dropna sounds sensible but what was the motivation you added it in this PR?

xiki-tempula:
@orbeckst So I would like to write a more robust gmx parser https://github.com/alchemistry/alchemlyb/pull/183.
However, I discovered that the new parser would fail some old tests. The problem seems to be associated with the load_expanded_ensemble_case_1, which when dropping the nan will result in a different answer.
So my question is would we expect NaN in the expanded ensemble dataset and are the NaNs artefacts?
I will try to contact the author of this dataset and see what is his opinion.

orbeckst:
@mrshirts we have a question about the [expanded ensemble dataset case 1](https://alchemtest.readthedocs.io/en/latest/gmx.html#alchemtest.gmx.load_expanded_ensemble_case_1): It contains NaN. Are they supposed to be in there?

mrshirts:
I don't think there should. Where specifically did it appear?  Which file/files?

orbeckst:
@xiki-tempula , you looked at the NaNs in the test data. Could you provide specific details to @mrshirts 's question, please? Thanks.

xiki-tempula:
@orbeckst Sorry, I'm a bit occupied on weekdays will have a look tomorrow.

xiki-tempula:
@mrshirts So there is only one file in gmx_expanded_ensemble_case_1, which is CB7_Guest3_dhdl.xvg
the line 1530, it is `2936.0000000000 21.0000000000 -91168.9690000000 nan -78.7414550000 113.0974300000 1.5736808000 64.8738410000 64.8738410000 64.8738410000 64.8738410000 64.8738410000 60.9368710000 56.9996240000 52.2751040000 47.5506900000 42.8262650000 38.1016370000 33.3771240000 28.6527460000 23.9282540000 17.6289880000 11.3295930000 5.0303196000 -1.2689984000 -7.5682849000 -13.8674930000 -9.6008139000 0.0000000000 12.5142880000 26.8151620000 39.1129320000 45.4792640000 58.5289710000 71.9205120000 85.6069280000 99.5163480000 113.6875200000 128.5488700000`, one could see the nan

orbeckst:
@mrshirts , after looking at the data https://github.com/alchemistry/alchemlyb/pull/184#issuecomment-1052074007 do you think we can simply decide to ignore this line in the data file?

mrshirts:
For now, yes, I think ignoring is best. I  haven't been able to figure out
why it is there, but we've used this data for other purposes, so I think it
is fine for demonstration.  We can eventually generate another data set as
necessary.

orbeckst (changes requested):
We need to adjust the tests. With the data we even have a builtin test case…

Please also add docs and CHANGELOG.

xiki-tempula:
@orbeckst Sorry, I opened this PR to show that there is NaN in the test set. It is not that I wish to merge this PR.

PR #188, opened 2022-03-05
Implement the base workflow class

xiki-tempula:


src/alchemlyb/tests/test_workflow.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Test_automatic_base():
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def workflow():
+        workflow = base.WorkflowBase()
+        workflow.run()
+        yield workflow
+
+    # def test_write(self, workflow, tmpdir):
```
xiki-tempula, line 15: @orbeckst Thanks for the suggestion of using pytest.patch and tmpdir. However, I found it a bit difficult to use it in this case.
I wish to change the workflow.out to the temporary directory using patch and then save the result to the temporary directory, where workflow is a pytest fixture. I wonder what is the best way of doing this? Thank you.

orbeckst: I would configure the output directory when you set up the fixture. If there's a kwarg then pass the tmpdir to the constructor. Otherwise, do something like
```python
workflow.out = tmpdir.strpath
```
although they now recommend using [tmp_path](https://docs.pytest.org/en/7.0.x/how-to/tmp_path.html) instead.

Normally I'd say use
```python
@pytest.fixture
def def workflow(tmp_path):
    outdir = tmp_path / "out"   # this out subdir is really optional, you can also just use tmp_path
    outdir.mkdir()              # and then you don't have to make the subdir either
    workflow = base.WorkflowBase(out=str(outdir))
    return workflow.run()
```
but that won't work with class-scoped fixtures.  There's a way to create these directories with [tmp_path_factory](https://docs.pytest.org/en/7.0.x/how-to/tmp_path.html#the-tmp-path-factory-fixture). I *think* you'd use it like this:
```python
@pytest.fixture(scope="session")
def def workflow(tmp_path_factory):
    outdir = tmp_path_factory.mktemp("out")
    workflow = base.WorkflowBase(out=str(outdir))
    return workflow.run()
```
According to the docs, it will create a session-scoped fixture. Therefore, you'd just put this fixture into the module instead of the class. I don't think that this would be a problem.

orbeckst (commented):
Taking on shape nicely, see comments inline to your question.

xiki-tempula:
@orbeckst Thanks. I have made the change.

src/alchemlyb/tests/test_workflow.py
```diff
@@ -0,0 +1,32 @@
...
+class Test_automatic_base():
+    @staticmethod
+    @pytest.fixture(scope='session')
+    def workflow(tmp_path_factory):
+        outdir = tmp_path_factory.mktemp("out")
+        workflow = base.WorkflowBase(out=str(outdir))
+        workflow.run()
+        return workflow
+
+    def test_write(self, workflow, tmpdir):
```
orbeckst, line 15: remove tmpdir, not used

src/alchemlyb/tests/test_workflow.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from alchemlyb.workflows import base
+import pandas as pd
+import os
+
+class Test_automatic_base():
+    @staticmethod
+    @pytest.fixture(scope='session')
```
orbeckst, line 8: session scope works inside a class? Probably because as a static method...

src/alchemlyb/workflows/base.py
```diff
@@ -0,0 +1,126 @@
...
+            The result of the convergence analysis.
+
+        """
+        self.convergence = pd.DataFrame()
+
+    def plot(self, *args, **kwargs):
+        """ The function for producing any plots.
+
+        """
+        ...
```
orbeckst, line 126: replace `...` with `pass`

orbeckst (changes requested):
Looks pretty good, just 
- minor things in comments
- update CHANGELOG

We might have to update docs and code as we go along but this looks good as the initial framework.

orbeckst:
Please ping me when you've finalized the PR & I'll merge it.

xiki-tempula:
@orbeckst Thanks for the review. I have made the changes. I have add the gmx parser to the change as well in case it gives conflict.

CHANGES
```diff
@@ -23,13 +23,13 @@ Changes
     (#171, PR #183)
 
 Enhancements
+  - Add a base class for the workflow.
```
orbeckst, line 26:
```suggestion
  - Add a base class for workflows (PR #188).
```

src/alchemlyb/workflows/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Attributes
+    ----------
+
+    file_list : list
+        A list of files to be read by the parser.
+
+    """
+    def __init__(self, units='kT', software='Gromacs', T=298, out='./', *args,
+    **kwargs):
```
orbeckst, line 31:
```suggestion
                 **kwargs):
```

src/alchemlyb/workflows/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+        self.T = T
+        self.software = software
+        self.unit = units
+        self.file_list = []
+        self.out = out
+
+    def run(self, *args, **kwargs):
+        """ Run the flow in an automatic fashion.
```
orbeckst, line 40:
```suggestion
        """ Run the workflow in an automatic fashion.
```

src/alchemlyb/workflows/base.py
```diff
@@ -0,0 +1,126 @@
...
+            The main result of the workflow.
+        convergence : pandas.Dataframe
+            The result of the convergence analysis.
+
+        """
+        self.read()
+        self.preprocess()
+        self.estimate()
+        self.check_convergence()
+        self.plot()
```
orbeckst, line 63: Would you not want to pass the args and kwargs to all intermediate stages?

docs/workflows/alchemlyb.workflows.base.rst
```diff
@@ -0,0 +1,27 @@
...
+step-by-step fashion. ::
+
+    >>> from alchemlyb.workflows.base import WorkflowBase
+    >>> workflow = WorkflowBase()
+    >>> workflow.read()
+    >>> workflow.preprocess()
+    >>> workflow.estimate()
+    >>> workflow.check_convergence()
+    >>> workflow.plot()
+
```
orbeckst, line 22: Add more text that describes the conventions that all workflow implementations should follow. For example:

* How are parameters passed? Is the convention that all methods take generic `*args, **kwargs` and then pick out the ones they need? 
* Which attributes are *required* in the class, if any? If it's the one documented in the base class then you can say so and link to the class docs.

orbeckst (changes requested):
Mainly clarification/doc questions.

(Sorry, I don't have much time at the moment, so very brief comments.)

docs/workflows/alchemlyb.workflows.base.rst
xiki-tempula: @orbeckst 
> How are parameters passed? Is the convention that all methods take generic *args, **kwargs and then pick out the ones they need?
Yes, I think "all methods take generic *args, **kwargs and then pick out the ones they need" might be the best initial solution.
>Which attributes are required in the class, if any? If it's the one documented in the base class then you can say so and link to the class docs.
So for the initialization, the creation of the class takes in units, software, T and the output directory. The documentation for this is created automatically by sphinx.

xiki-tempula:
@orbeckst Thanks for the review. I have made the changes.

docs/workflows/alchemlyb.workflows.base.rst
orbeckst: Say explicitly what is required in text. Textbooks important to make Intention clear. You’re defining the API here so make it easy for other developers by pointing out what they need to consider. Just „look at existing code“ ist not developer friendly.

orbeckst: Also say clearly that the methods will have to take args, kwargs.

orbeckst (changes requested):
Please write more text, mainly for developers. When defining a new API it’s important to be clear about what is required and what is optional. Think about other developers wanting to contribute code. The better you describe now what they should do, the less you have to answer questions later, and the more consistent the library will be for users. 

Otherwise all good!!

xiki-tempula:
@orbeckst Thank you for the comment. I wonder if there is any examples that I could follow?

xiki-tempula:
@orbeckst Thanks for the coments. I have added more text to describe the initialisation method and the run method.

orbeckst (approved):
very nice, the added docs are really useful

orbeckst:
Hooray, well done @xiki-tempula , we now have the workflows module!! 🎉

PR #190, opened 2022-03-22
Add autombar to the convergence analysis list

### LINKED ISSUE #189: Support AutoMBAR in forward and backward convergence
I currently have a data set which can only be analyzed by autombar (related to one issue in pymber https://github.com/choderalab/pymbar/issues/419 ). I want to get the forward and backward convergence plot using mbar, and I found the current version doesn't support autombar in convergence analysis. 
It would help if it's supported.

xiki-tempula:
Fix #189


orbeckst (approved):
lgtm, thank you @xiki-tempula !!

PR #191, opened 2022-04-10
adopt pymbar 3.0.6 and conda for CI

xiki-tempula:


orbeckst:
It looks that with conda it takes about 11 min for the tests and with the previous pip-based between 9min and 11 mins.

Are you already using mamba?

xiki-tempula:
@orbeckst Are you already using mamba could indeed speed the CI up for linux and osx but it won't work for the windows. I could create another action for windows which uses conda.

orbeckst:
I looked at the 3.7 windows action and the alchemlyb build step fails with a weird "conda command not found". Do you know why?

xiki-tempula:
@orbeckst Sorry, my fault. Should work now.

orbeckst:
CI is hanging and I wasn't able to restart it. Need to check back later.

orbeckst:
All of the window runners hang reproducibly in the tests — the last time at the plot mbar overlap.

Do you have a local Windows machine where you could test this?

xiki-tempula:
@orbeckst Ok, I think it is fixed now.

src/alchemlyb/tests/test_visualisation.py
```diff
@@ -1,4 +1,5 @@
 import matplotlib
+matplotlib.use('Agg')
```
orbeckst, line 2: Instead of hard coding in the test, set an environment variable in the ci.yml file
```yaml
    env:
      MPLBACKEND: agg
```

devtools/conda-envs/test_env.yaml
```diff
@@ -0,0 +1,16 @@
+name: test
+channels:
+- conda-forge
+dependencies:
+- python
```
orbeckst, line 5: Do we need to specific python here? Does this work with the Python matrix?

src/alchemlyb/tests/test_visualisation.py
orbeckst: P.S.: Good that you identified the issue!

orbeckst (changes requested):
Looks pretty good, I'd just use an env variable for agg.

devtools/conda-envs/test_env.yaml
xiki-tempula: It is here
[extra-specs: |
            python=${{ matrix.python-version }}](https://github.com/alchemistry/alchemlyb/pull/191/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR36)

orbeckst: My English was bad: What I meant was "Do we need to include (=specify) python in this file at all as a dependency, given that we are already setting it in the build matrix?" 

Will the CI also work if there's no "python" inside test_env.yaml? But that's a minor thing, given that CI works at the moment.

.github/workflows/ci.yaml
```diff
@@ -19,44 +19,43 @@ jobs:
...
+            pytest=${{ matrix.pytest }}
 
     - name: Install alchemtest
+      # conda setup requires this special shell
+      shell: bash -l {0}
       run: |
         python -m pip install https://github.com/alchemistry/alchemtest/archive/master.zip
 
-    - name: Install package (with remaining dependencis)
+    - name: Install package (with no dependencis)
```
orbeckst, line 45:
```suggestion
    - name: Install package (with no dependencies)
```

orbeckst (approved):
Awesome, the use of µmamba reduced our overall CI run time. Nice work!!

PR #194, opened 2022-05-26
Set a strict python minimum version for install, add automated pypi deployment

### LINKED ISSUE #193: Minimum python version isn't being enforced
The way alchemlyb's setup files are currently setup means that you can install it using an unsupported Python version.
(@orbeckst same issue as https://github.com/MDAnalysis/GridDataFormats/issues/112)

This primarily affects the the 0.6.0 release on pypi where you have the following two issues:
1. Universal wheels set as py2-py3
2. It can be installed without any issues with py36 even though the release notes claim >py37.

It is worth noting that it doesn't currently affect the `master` branch, but that's purely because the pandas requirements can't be met. You end up getting a `no matching distribution found for pandas` error rather than being told that the issue is down to your python version.

The fix here is rather simple:
1. Add a python_requires to setup.py
2. Remove the building of universal wheels in setup.cfg

I'm also happy to implement the automated wheel generation actions we implemented in GridDataFormats whilst @orbeckst and I were fixing this issue, if it's of interest.

IAlibay:
Fixes #193 

Work done in this PR:
  - two small changes to setup.py and setup.cfg to enforce python>=3.7 and prevent universal wheels (which are meant to also work with py2) from being built.
  - Added automatic deployment to pypi on release, as per: https://github.com/MDAnalysis/GridDataFormats/blob/master/.github/workflows/deploy.yaml (I'm happy saying that bit of code is under BSD-3 if that works with you @orbeckst)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,6 @@ jobs:
         environment-name: test
         extra-specs: |
             python=${{ matrix.python-version }}
-            pytest=${{ matrix.pytest }}
 
     - name: Install alchemtest
       run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,9 +11,16 @@ on:
     branches:
       - "master"
 
+
 concurrency:
-  group: "${{ github.ref }}-${{ github.head_ref }}"
+  group: "${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}"
   cancel-in-progress: true
+
+
+defaults:
+  run:
+    shell: bash -l {0}
+
 
 jobs:
   test:
@@ -25,7 +32,7 @@ jobs:
         python-version: ["3.7", "3.8", "3.9", "3.10"]   
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     # More info on options: https://github.com/conda-incubator/setup-miniconda
     - uses: mamba-org/provision-with-micromamba@main
@@ -37,28 +44,21 @@ jobs:
             pytest=${{ matrix.pytest }}
 
     - name: Install alchemtest
-      # conda setup requires this special shell
-      shell: bash -l {0}
       run: |
         python -m pip install https://github.com/alchemistry/alchemtest/archive/master.zip
 
     - name: Install package (with no dependencies)
-      # conda setup requires this special shell
-      shell: bash -l {0}
       run: |
         python -m pip install . --no-deps
 
     - name: Run tests
-      # conda setup requires this special shell
-      shell: bash -l {0}
-
       run: |
         pytest -v --cov=alchemlyb --cov-report=xml --color=yes src/alchemlyb/tests
       env:
         MPLBACKEND: agg
 
     - name: Codecov
-      uses: codecov/codecov-action@v1.5.2
+      uses: codecov/codecov-action@v2
       with:
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
         file: ./coverage.xml
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,9 +1,6 @@
 name: Build and upload to PyPI
 
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
       - master
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -26,7 +26,7 @@ defaults:
 
 jobs:
   build_wheels:
-    environment: deploy
+    # environment: deploy
     if: "github.repository == 'alchemistry/alchemlyb'"
     name: Build pure Python wheel and tarball
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


orbeckst:
work for me, it's your code!

.github/workflows/ci.yaml
```diff
@@ -11,10 +11,17 @@ on:
     branches:
       - "master"
 
+
 concurrency:
-  group: "${{ github.ref }}-${{ github.head_ref }}"
+  group: "${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}"
```
IAlibay, line 16: This will be necessary to avoid push clashes with the other yaml file.

.github/workflows/ci.yaml
```diff
@@ -11,10 +11,17 @@ on:
...
   cancel-in-progress: true
 
+
+defaults:
```
IAlibay, line 20: This is more of a suggested change (the old way isn't wrong, I just see this as cleaner but I'm happy to revert)

.github/workflows/ci.yaml
```diff
@@ -37,28 +44,21 @@ jobs:
...
-      shell: bash -l {0}
-
       run: |
         pytest -v --cov=alchemlyb --cov-report=xml --color=yes src/alchemlyb/tests
       env:
         MPLBACKEND: agg
 
     - name: Codecov
-      uses: codecov/codecov-action@v1.5.2
+      uses: codecov/codecov-action@v2
```
IAlibay, line 61: v1 is deprecated as far as I remember

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,67 @@
+name: Build and upload to PyPI
+
+on:
+  pull_requests:
+    branches:
+      - master
```
IAlibay, line 6: This will get removed once it works on CI in this PR.

For things to work you'll need to upload two tokens to github secrets (I've set them to exist within an environment named `deploy`, but you can make them global secrets too if you'd like).
1. TEST_PYPI_API_TOKEN -> the API token for the test.pypi package of alchemlyb
2. PYPI_API_TOKEN -> the API token for the pypi package of alchemlyb

I recommend having the testpypi check there (so you can tell if something went wrong before pushing to pypi), but it's not a necessity either.

orbeckst: I added these two tokens to the *deploy* environment.

orbeckst:
Yes please on doc update. 

Thank you!

Btw, did you add yourself to changelog? … not sure if we have AUTHORS.

IAlibay:
Doesn't look like I can edit the wiki unfortunately (maybe a limitation to require at least one commit?), authors and changelog have been updated.

Here is the wiki text to modify if you would like to do it:

```
* Releases are cut from the master branch and tagged with *MAJOR.MINOR.PATCH* (note: the release tag *determines* the tag because we use [versioneer](https://github.com/warner/python-versioneer/blob/master/INSTALL.md#post-installation-usage), which obtains the release number from the git tag). We do from the master branch for the **PyPI release**
    1. `git tag <major>.<minor>.<patch>`
    2. `git push --tags`

* Note: the tags need to be pushed to GitHub with `git push --tags`. (You can add text to the [releases on GitHub](https://github.com/alchemistry/alchemlyb/releases).)

* This will automatically trigger a github action (named `deploy`) to use pypa's `build` tool to create a tarball and & pure Python wheel and upload it on [testpypi](https://test.pypi.org/project/alchemlyb)
* Once uploaded to testpypi, please check locally that the build is working as intended. In a clean environment do:
    1. `pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple alchemlyb=="version number"`
    2. `pip install pytest`
    3. `pip install https://github.com/alchemistry/alchemtest/archive/master.zip`
    4. `pytest --pyargs alchemlyb`

* After this, create a [release on GitHub](https://github.com/alchemistry/alchemlyb/releases); zenodo DOIs and snapshots are automatically minted.
* The release will trigger the `deploy` action again, which will this time upload the tarball and wheel to [PyPI](https://pypi.org/project/alchemlyb/).
```

xiki-tempula:
@IAlibay Many thanks for the PR, is it possible to have automatic conda deployment as well? So we can have everything automated.

IAlibay:
> @IAlibay Many thanks for the PR, is it possible to have automatic conda deployment as well? So we can have everything automated.

conda-forge should be automatically picking up from pypi as it is, the bot will automatically trigger a PR. You can set cf-bot to auto merge PRs but I don't recommend it.

There are wheels for pushing wheels directly but honestly my 2 cent is that the current method in place is just saner, especially if you alchemlyb moves away from being pure python one day.

xiki-tempula (approved):
LGTM

.github/workflows/ci.yaml
orbeckst: thanks

.github/workflows/ci.yaml
orbeckst: nicer, thanks!

orbeckst (approved):
Awesome, thank you so much @IAlibay !

I updated the wiki https://github.com/alchemistry/alchemlyb/wiki/Developer-Guide#release-management

orbeckst:
Thank you @IAlibay ! Sorry, took me a while to look at it again — thank you @xiki-tempula for the reminder.

PR #199, opened 2022-06-22
Sync series and dataframe in statistical_inefficieny()

### LINKED ISSUE #198: statistical_inefficiency does not slice data frame
### Description

When using `alchemlyb.preprocessing.statistical_inefficiency` with a data frame, a series to calculate statistical inefficiency on, and slicing parameters (`lower`, `upper`, `step`), the slicing is only applied to the series, but not to the data frame. Since `statistical_inefficiency` uses a raw index to subsample the data frame, the series and the data frame are out of sync by the end of the function. This leads to results that seem, at the very least, unintuitive.

### Examples
I think that a few examples make this much easier to explain. Assume I have loaded some data, e.g.
```python
data = extract_u_nk('lambda_0.xvg', T=298.15)
```
and want to use only part of this data, e.g.
```python
lower = 1000
upper = 5000
```
then I would expect a few things:

1. Slicing removes times before and after my set limits → this works ✅ 
```python
from alchemlyb.preprocessing import slicing
sliced_data = slicing(data, lower=lower, upper=upper)
assert all(sliced_data.reset_index()['time'] >= lower)
assert all(sliced_data.reset_index()['time'] <= upper)
```

2. Using the slicing arguments with `statistical_inefficiency` removes times before and after my set limits → this does not work ❌  
```python
from alchemlyb.preprocessing import statistical_inefficiency
subsampled_data = statistical_inefficiency(data, data.columns[0], lower=lower, upper=upper)
assert all(subsampled_data.reset_index()['time'] >= lower)
assert all(subsampled_data.reset_index()['time'] <= upper)
```
Specifically, I find that the times are shifted, i.e. that the lowest time of `subsampled_data` is 0, and the highest time is `upper - lower`.

3. Using slicing on the data frame before running `statistical_inefficiency`, or running `statistical_inefficiency` with slicing parameters leads to the same result → this does not work ❌ 
```python
sliced_data = slicing(data, lower=lower, upper=upper)
subsampled_sliced_data = statistical_inefficiency(sliced_data, data.columns[0])

subsampled_data = statistical_inefficiency(data, data.columns[0], lower=lower, upper=upper)

assert (subsampled_data == subsampled_sliced_data).all(axis=None)
```
The problem here is, unsurprisingly, similar to the one above: `subsampled_data` starts at time 0, ends at time `upper - lower`, whereas `subsampled_sliced_data` starts at time `lower` and ends at time `upper`.

4. The issues demonstrated in cases 2 and 3 are especially problematic if we imagine a use case in which we would repeatedly sample from the same data, e.g.
```python
window_size=1000
for window_idx in range(5):
    lower = window_idx * window_size
    upper = (window_idx + 1) * window_size
    subsampled_data = statistical_inefficiency(data, data.columns[0], lower=lower, upper=upper)
    # ...
    # further analysis of subsampled_data for window here
    # ...
```

My expectation here would be to analyze windows of different data at every iteration. The current implementation would, however, always subsample from the window 0 to 1000. (The subsampled data might differ between windows, since the statistical inefficiency is actually calculated on a different window of the series each time, but the resulting indices are applied to the same window of the data frame every time.)

5. These issues could lead to correlated samples in case that the `step` keyword is used. Imagine that we have data which is correlated such that the `conservative` subsampling algorithm picks every 10th frame. If we were to use `statistical_inefficiency` with `step=10` (by chance, or because we have a feeling that we might have sampled data too frequently), then `statistical_inefficiency` would return every frame of the first 10% of the data! So we would have lost a lot of information, and would continue working with heavily sampled data.

### Proposed solution
I will propose a change shortly which adds tests (in line with examples 1 to 3) and a fix for this behavior. I think the most stable fix is to call `slicing` on both the series and the dataframe. It might be possible to calculate an offset index plus a stride to avoid slicing the dataframe, but that seems like pointless overengineering since there is a `slicing` function readily available that makes sure that the series and dataframe slicing will always be done in the same way.

PS: Maybe I've misunderstood everything and this is intended behavior. In this case, please close the PR, but I would then suggest to add some documentation (ideally also a note when calling the function) warning about this behavior 🙂

ptmerz:
This adds a number of tests for slicing() and statistical_inefficieny():
* Test that slicing() respects upper and lower time bounds (currently
  passes)
* Test that statistical_inefficieny() respects upper and lower time bounds
  when it is used without series to subsample (currently passes)
* Test that statistical_inefficieny() respects upper and lower time bounds
  when it is used with series to subsample (currently fails)
* Test that first using slicing() on the data frame, then
  statistical_inefficieny() without time bounds yields the same results as
  a single call to statistical_inefficieny() with time bounds (currently
  fails)

It then makes sure that both the subsampling series and the dataframe are sliced
when calling statistical_inefficieny(). This ensure that the series and
the dataframe are in sync before the subsampled indices of the series
are applied to the dataframe. This fixes the above tests.

Fixes #198


src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -68,6 +68,29 @@ def slicer(self, *args, **kwargs):
...
+        assert any(data.reset_index()['time'] > upper)
+
+        # Slice data, and check that we don't observe times outside
+        # the prescribed range
+        sliced = self.slicer(data, lower=lower, upper=upper, step=5)
+        assert all(sliced.reset_index()['time'] >= lower)
+        assert all(sliced.reset_index()['time'] <= upper)
+
+        # Make sure we didn't change input data
+        assert len(data) == original_length
```
xiki-tempula, line 92: Why do you need to check if the input data changes in the unit test?

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -213,6 +236,97 @@ def test_raise_ValueError_for_mismatched_data(self, series):
...
+        """
+        original_length = len(data)
+        # Check that the input data is appropriate for the test
+        assert any(data.reset_index()['time'] < lower)
+        assert any(data.reset_index()['time'] > upper)
+
+        # Slice data, and check that we don't observe times outside
+        # the prescribed range
+        sliced = self.slicer(data,
+                             series=None,
```
xiki-tempula, line 257: Sorry, I didn't quite see how is this different from `test_lower_and_upper_bound` as the series is default to None?

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -213,6 +236,97 @@ def test_raise_ValueError_for_mismatched_data(self, series):
...
+                                 (gmx_benzene_u_nk(), 1000, 34000),
+                             ])
+    def test_lower_and_upper_bound_inefficiency(self, data, lower, upper):
+        """
+        Test that the lower and upper time is respected when using statistical_inefficiency
+        with a series. In this case, statistical_inefficiency should slice the series, then
+        subsample the data frame.
+        """
+        original_length = len(data)
+        # Check that the input data is appropriate for the test
```
xiki-tempula, line 279: I think you don't need this for every tests.
Might create a test called `test_sanity` and check for this.

xiki-tempula (changes requested):
Many thanks for the contribution. This fix is very important.
I think the test could be simplified as bit using the @pytest.mark.parametrize

You could do something like
```python
@pytest.mark.parametrize('data', (gmx_benzene_dHdl(), gmx_benzene_u_nk())
@pytest.mark.parametrize('test_series', (True, False))
def test_lower_and_upper_bound_slicer(self, data, test_series):
        """
        Test that the lower and upper time is respected when using statistical_inefficiency
        without a series. In this case, statistical_inefficiency should behave like slicing
        """
        lower, upper = 1000, 34000
        original_length = len(data)
        # Check that the input data is appropriate for the test
        assert any(data.reset_index()['time'] < lower)
        assert any(data.reset_index()['time'] > upper)

        if test_series:
           series=data.sum(axis=1)
        else:
           series=None

        # Slice data, and check that we don't observe times outside
        # the prescribed range
        sliced = self.slicer(data,
                             series= series,
                             lower=lower,
                             upper=upper,
                             step=5)
        assert all(sliced.reset_index()['time'] >= lower)
        assert all(sliced.reset_index()['time'] <= upper)

        # Make sure we didn't change input data
        assert len(data) == original_length
```

src/alchemlyb/tests/test_preprocessing.py
ptmerz: I think it's valuable to check that slicing does not change the input data. However, this might be cleaner as its own unit test.

Shall I implement this? Essentially, I would duplicate the entire test until the call to the slicer, then assert only that the original data is unchanged, and call it something like `test_preserves_original_data`.

Or, we could change the name of the current test to `test_lower_and_upper_bound_and_preserves_original_data` to reflect that we are testing two things.

Which approach would you suggest?

src/alchemlyb/tests/test_preprocessing.py
ptmerz: Well, this test class tests `statistical_inefficiency(...)`, the one above tests `slicing(...)`.

So this tests makes sure that `statistical_inefficiency` respects the upper and lower bounds when it's used without series. The test above makes sure that `slicing` respects the upper and lower bounds.

This test could be rewritten as equivalence between `statistical_inefficiency` and `slicing`, if you prefer.

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -213,6 +236,97 @@ def test_raise_ValueError_for_mismatched_data(self, series):
...
+        sliced = self.slicer(data,
+                             series=None,
+                             lower=lower,
+                             upper=upper,
+                             step=5)
+        assert all(sliced.reset_index()['time'] >= lower)
+        assert all(sliced.reset_index()['time'] <= upper)
+
+        # Make sure we didn't change input data
+        assert len(data) == original_length
```
ptmerz, line 265: Note: We should treat this in the same way as the test above (see comment line 92).

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -213,6 +236,97 @@ def test_raise_ValueError_for_mismatched_data(self, series):
...
+        # the prescribed range
+        sliced = self.slicer(data,
+                             series=data.sum(axis=1),
+                             lower=lower,
+                             upper=upper,
+                             step=5)
+        assert all(sliced.reset_index()['time'] >= lower)
+        assert all(sliced.reset_index()['time'] <= upper)
+        # Make sure we didn't change input data
+        assert len(data) == original_length
```
ptmerz, line 293: Note: We should treat this in the same way as the test above (see comment line 92).

src/alchemlyb/tests/test_preprocessing.py
ptmerz: Here's my rationale for implementing this on every test:

There is at least three ways to render this test useless:
* Someone changes the input data to this test from `gmx_benzene_*()` to something else
* Someone changes the lower / upper bound used in this test
* Someone changes the underlying data that `gmx_benzene_*()` returns

The solution you suggest addresses my third point. But then you'd need a check to make sure that the two tests are using the same input data...

Additionally, this check is also an information for developers looking at this test - we are clearly stating what we are expecting coming in, so the rest of the test is easier to understand.

ptmerz:
> I think the test could be simplified as bit using the @pytest.mark.parametrize

Agreed that this could unify the second and third added test (`TestStatisticalInefficiency. test_lower_and_upper_bound_slicer` and `TestStatisticalInefficiency.test_lower_and_upper_bound_inefficiency`). Unless we decide to rewrite `TestStatisticalInefficiency. test_lower_and_upper_bound_slicer` as an equivalence between `statistical_inefficiency` and `slicing`, of course.

So I'd suggest to wait until we have agreed on how to proceed with the remaining points raised, and I will implement this simplification if it's still viable!

src/alchemlyb/tests/test_preprocessing.py
orbeckst: Separate test would be cleaner, especially once it fails. I think it's sufficient to check this with just a single input data set. You probably also don't need the check that the data are appropriate for the test.

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -68,6 +68,29 @@ def slicer(self, *args, **kwargs):
     def test_basic_slicing(self, data, size):
         assert len(self.slicer(data, lower=1000, upper=34000, step=5)) == size
 
+    @pytest.mark.parametrize(('data', 'lower', 'upper'),
+                             [
+                                 (gmx_benzene_dHdl(), 1000, 34000),
+                                 (gmx_benzene_u_nk(), 1000, 34000),
```
orbeckst, line 74: It's not great to call the data provider functions here because this adds to the time during which tests are gathered in serial as opposed during test execution (which can be parallelized). I know that we already have plenty examples of doing exactly that elsewhere so it's really not your fault.

I am just wondering if there are better ways to do this, e.g., passing the function itself as `data_generator` and then calling it as `data = data_generator()` might be easiest. It would be better if we could generate them as fixtures (but parametrized fixtures are still awkward IIRC).

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -213,6 +236,97 @@ def test_raise_ValueError_for_mismatched_data(self, series):
...
+                             ])
+    def test_lower_and_upper_bound_slicer(self, data, lower, upper):
+        """
+        Test that the lower and upper time is respected when using statistical_inefficiency
+        without a series. In this case, statistical_inefficiency should behave like slicing
+        """
+        original_length = len(data)
+        # Check that the input data is appropriate for the test
+        assert any(data.reset_index()['time'] < lower)
+        assert any(data.reset_index()['time'] > upper)
```
orbeckst, line 252: I am not 100% convinced that this is necessary, given that you are the one choosing the test data. It guards against the case when someone else adds other data to the parameters, so I am not totally opposed to having it here and elsewhere. However, in order to reduce code duplication, can you make it a small function such as `_assert_data_has_boundaries(data, lowe, upper)` or rather choose a better name.

orbeckst: I just saw your [explanation for why you include these guards everywhere](https://github.com/alchemistry/alchemlyb/pull/199/files#r909367088). I am ok with leaving it, I'd just reduce copy&paste code.

src/alchemlyb/tests/test_preprocessing.py
orbeckst: makes sense; I'd repackage as a function (see my other comment)

orbeckst (changes requested):
Overall, looks good to me — just small comments on the tests.

src/alchemlyb/tests/test_preprocessing.py
xiki-tempula: @orbeckst The parametrized fixtures could be made easy with `pytest-lazy-fixture`

I have posted an example here as is noted in their website https://pypi.org/project/pytest-lazy-fixture/. This is conda installable as well.
```
@pytest.fixture(params=[
    pytest.lazy_fixture('one'),
    pytest.lazy_fixture('two')
])
def some(request):
    return request.param

@pytest.fixture
def one():
    return 1

@pytest.fixture
def two():
    return 2
```

src/alchemlyb/tests/test_preprocessing.py
xiki-tempula: @ptmerz  Why do you need to check slicing when you are changing the code in statistical_inefficiency?

src/alchemlyb/tests/test_preprocessing.py
xiki-tempula: @ptmerz 

> Someone changes the underlying data that gmx_benzene_*() returns

I do agree with this one though it might be very less likely. a separate sanity test should be good enough.

> Someone changes the lower / upper bound used in this test

Do you mind elaborating on this? I think in general there is no need to change the test unless it failed but that would be a different case.

> Someone changes the input data to this test from gmx_benzene_*() to something else
I don't think you need to guard this as I cannot think of a user case where this would happen?

src/alchemlyb/tests/test_preprocessing.py
xiki-tempula: @ptmerz It might be good to abstract out the core of these three tests out as fixture and use parametrize to go through them

something like
@pytest.mark.parametrize(
    'output_data,original_length',
    [
        (lazy_fixture('inefficiency'), 100),
        (lazy_fixture('slicing'), 1000),
    ]
def test_original_length(output_data, original_length):
    assert len(data) == original_length

src/alchemlyb/tests/test_preprocessing.py
ptmerz: I don't _need_ to. While investigating the bug I reported I noticed that there was no test for ranges in `slicing` or in `statistical_inefficiency`, which made it possible for the bug in `statistical_inefficiency` to go unnoticed until now. I thought it would be helpful to broaden the test coverage also for `slicing` to avoid any future bugs. Do you prefer me to remove them?

src/alchemlyb/tests/test_preprocessing.py
ptmerz: > @ptmerz
> 
> > Someone changes the underlying data that gmx_benzene_*() returns
> 
> I do agree with this one though it might be very less likely. a separate sanity test should be good enough.

Sure, that's the case that would be covered by a separate test.

> > Someone changes the lower / upper bound used in this test
> 
> Do you mind elaborating on this? I think in general there is no need to change the test unless it failed but that would be a different case.

I am not referring to needing to changing the tests. My guiding principle when writing tests is to make them as secure as possible. Bugs are usually not introduced on purpose, but because someone misunderstood what a functionality (or a test) does, or someone was not aware of side effects that a change has in other parts of the code. This `assert` guards from someone mistakenly changing the input data of the test to something that stops this test from being useful.

> > Someone changes the input data to this test from gmx_benzene_*() to something else
> 
> I don't think you need to guard this as I cannot think of a user case where this would happen?

Again, that's not a user case. It could very well happen that in a while a developer comes up with a great new input data set to be used in tests. With the test implemented as I am suggesting, that person can confidently change the input data without having to understand the details of the test. Imagine we didn't have these additional tests. They might want to add shorter input data to speed up the run time of tests. They could add that and the test would still pass - after all, `assert all(sliced.reset_index()['time'] <= upper)` is going to always pass if the input data was shorter than the upper bound from the start.

ptmerz: Maybe some of the confusion comes from the lack of syntactic difference between checks to assert that certain assumptions that I (as test writer) made are correct, and the actual test of functionality. In natural language, this would go something like that:

**Given** a dataset and upper and lower bounds,  
**assuming that** the input data set has data that is lower than the lower bound and higher than the upper bound,  
**test that** `slicing` does successfully remove frames before the lower bound and after the higher bound, and does not change the input file.

The important part is that the test is useless if the assumptions don't hold. I am not making any guesses about how probable it is that someone changes the test files in a way that these assumptions fail.

Another way to look at it is from the point of view of the function. The test function is principally a function that says "I will assert that `slicing` works correctly". There's two ways to interpret this:
1. I will assert that `slicing` works if and only if you give me reasonable input data. Otherwise, my behavior is undefined (because I might still tell you that everything works although I haven't actually been able to check anything). 
2. I will assert that `slicing` works or loudly complain if I'm unable to perform that task with the input data you gave me.

I prefer the second approach, so that's what I went for.

src/alchemlyb/tests/test_preprocessing.py
ptmerz: Extracted this into separate test.

src/alchemlyb/tests/test_preprocessing.py
ptmerz: Done in latest version.

src/alchemlyb/tests/test_preprocessing.py
ptmerz: This makes a lot of sense @orbeckst, I wasn't aware of this issue. Unfortunately I haven't had time to look into this yet. Current version therefore still has the greedy evaluation.

orbeckst: `lazy_fixture` looks like a convenient solution. I looked at the package and I am not sure how well maintained it is. There's a commit from Jan 2022 in https://github.com/tvorog/pytest-lazy-fixture but the pypi page https://pypi.org/project/pytest-lazy-fixture/ states that the highest supported Python version is 3.7. @xiki-tempula , do you have experience with it in production code? It would be annoying to rewrite quite a few of our tests, just to discover that it breaks with a new release of pytest...

orbeckst: I had a look around in our tests and we do have examples of evaluating the loading functions later: see e.g. https://github.com/alchemistry/alchemlyb/blob/72622c99c16c74efd1f8517cb15ef21e16d6ebda/src/alchemlyb/tests/test_fep_estimators.py#L145-L168

This is not too bad.

orbeckst (changes requested):
Thanks a lot for all the work @ptmerz and explaining your choices to us — very much appreciated!

After looking at the up-front loading I'd ask you to change your code to load during run time and not set up. I posted an example from our code that should make it relatively straightforward to implement.

I will raise an issue to eventually change the other parts of the tests in a similar fashion. 

Please also add yourself to AUTHORS and add an entry to CHANGES.

Thank you!!

src/alchemlyb/tests/test_preprocessing.py
orbeckst: I raised #206 for fixing the other instances of upfront data loading.

xiki-tempula: @orbeckst yes, I used the lazy_fixture in _our_ production code but the reason that I use it is that it was there. It works fine for me but I do share your concern about it not being updated for years.

On the other hand, one could also use the official pytest solution

```
@pytest.fixture(params=['dir1_fixture', 'dir2_fixture'])
def dirname(request):
    return request.getfixturevalue(request.param)
```

orbeckst: This looks straightforward
```python
@pytest.mark.parametrize(('dataloader', 'size'), [('gmx_benzene_dHdl', 661), 
                                                  ('gmx_benzene_u_nk', 661)]) 
     def test_basic_slicing(self, dataloader, size, request):
         data = request.getfixturevalue(dataloader)
         assert len(self.slicer(data, lower=1000, upper=34000, step=5)) == size 
```

ptmerz:
@orbeckst , I uploaded new commits that should hopefully address your points. Note that currently, the data loading functions had to be duplicated, as they can't be normal functions and fixtures at the same time. If after resolving #206, you only use them as fixtures, you could remove this duplication.

CHANGES
```diff
@@ -14,7 +14,7 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 
-*/*/2022 xiki-tempula, IAlibay, dotsdl, orbeckst
+*/*/2022 ptmerz, xiki-tempula, IAlibay, dotsdl, orbeckst
```
orbeckst, line 17: It's customary that you append the name.
```suggestion
*/*/2022 xiki-tempula, IAlibay, dotsdl, orbeckst, ptmerz
```

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -241,6 +241,11 @@ def statistical_inefficiency(df, series=None, lower=None, upper=None, step=None,
        inefficiency was _rounded_ (instead of ``ceil()``) and thus one could
        end up with correlated data.
 
+    .. versionchanged:: 0.7.0
+       Fixed a bug that would effectively ignore the ``lower`` and ``step``
+       keywords when returning the subsampled DataFrame object. See
+       https://github.com/alchemistry/alchemlyb/issues/198 for more details.
```
orbeckst, line 247: Mark-up links
```suggestion
       `issue #198 <https://github.com/alchemistry/alchemlyb/issues/198>`_ for 
       more details.
```

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -15,11 +15,14 @@
...
 
-
 def gmx_benzene_dHdl():
     dataset = alchemtest.gmx.load_benzene()
     return gmx.extract_dHdl(dataset['data']['Coulomb'][0], T=300)
 
+@pytest.fixture()
+def gmx_benzene_dHdl_fixture():
+    dataset = alchemtest.gmx.load_benzene()
+    return gmx.extract_dHdl(dataset['data']['Coulomb'][0], T=300)
```
orbeckst, line 25: reduce code duplication and make clear that function and fixture are doing exactly the same thing
```suggestion
    return gmx_benzene_dHdl()
```

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -15,11 +15,14 @@
 
 import alchemtest.gmx
 
-
 def gmx_benzene_dHdl():
     dataset = alchemtest.gmx.load_benzene()
     return gmx.extract_dHdl(dataset['data']['Coulomb'][0], T=300)
 
+@pytest.fixture()
```
orbeckst, line 22:
```suggestion
# When issue #206 is addressed make the gmx_benzene_dHdl() function the 
# fixture, remove the wrapper below, and replace 
# gmx_benzene_dHdl_fixture --> gmx_benzene_dHdl
@pytest.fixture()
```

orbeckst (approved):
I have some small changes that I will apply directly. Everything looks really good!

One source code line change, lots of other stuff around it ... but that's what makes the difference for a well-tested library. Thank you for your willingness to see the whole PR through to the end! We very much appreciated your contribution.

PR #214, opened 2022-08-14
remove 3.7 support

xiki-tempula:
Remove the support for py3.7


environment.yml
```diff
@@ -5,7 +5,7 @@ dependencies:
 - python
 - numpy
 - pandas
-- pymbar
+- pymbar >=3.0.5,<4
```
orbeckst, line 8: oops, ... must have forgotten this earlier

src/alchemlyb/__init__.py
```diff
@@ -46,15 +46,6 @@ def concat(objs, *args, **kwargs):
...
-        attrs = objs[0].attrs
-    except IndexError: # except empty list as input
-        raise ValueError('No objects to concatenate')
-
-    for obj in objs:
-        if attrs != obj.attrs:
-            raise ValueError('All pandas objects should have the same attrs.')
-    new = pd.concat(objs, *args, **kwargs)
-    new.attrs = attrs
-    return new
```
orbeckst, line 60: keep the original code, don't replace with pd.concat

orbeckst: (In another discussion on MDAnalysis, @IAlibay convinced me that we should just preserve the deprecated code exactly as is; also minimizes the diff)

src/alchemlyb/__init__.py
```diff
@@ -46,15 +46,6 @@ def concat(objs, *args, **kwargs):
...
+    warn('This method will be deprecated in 1.0.0. Use pd.concat instead.',
```
orbeckst, line 49: Fix text:

This function will me removed in 1.0.0. Use pandas.concat() instead.

readthedocs.yml
```diff
@@ -8,7 +8,7 @@ formats:
   - pdf
 
 python:
-  version: 3.7
+  version: 3.8
```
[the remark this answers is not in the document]
xiki-tempula: Error
Problem in your project's configuration. Invalid "python.version": expected one of (2, 2.7, 3, 3.5, 3.6, 3.7, 3.8, pypy3.5), got 3.9

xiki-tempula:
@orbeckst 
I'm also a bit reluctant to remove the check and we shall keep the alchemlyb.concat.
I'm happy to drop the support of py3.7.

orbeckst (changes requested):
please add CHANGELOG entry, otherwise looking good

xiki-tempula:
@orbeckst Done.

orbeckst (approved):

PR #219, opened 2022-09-01
issue-#221-Amber parser incorrectly set the time when reading dHdL/ u_nk values

DrDomenicoMarson:
Hello everyone, this is the first time for me contributing to anything on GitHub, I hope I haven't done anything wrong 😄 
I think how the current parser computes the time value for each value of dHdL/u_nk is wrong. Currently, the logic is like this:
```
file_datum.t0 + (frame_index + 1) * file_datum.dt * 1000
```
where the ```1000```is arbitrary as far as I understand (it's stated that is a unit conversion, but that's not true).

This is not a big deal when we are reading just from a single file, because the time index will monotonically increase anyway, but problems arise when we try to concatenate multiple dataframes coming from continuations from the same simulation. What's happening is that every time a new output file is read, the "time origin" is read from the file itself, and so it's reset, and from here the time is correctly "increased". Concatenating these files will easily produce wrong dataframes with incorrect/duplicate time indexes. 

It happened to me, that I had simulations with dt=0.002 ps, ntpr=500 steps, and restarting every 5 ns, so the output file from every simulation had 5000 values of dHdL/u_nk. When importing two consecutive output files:
- the first df had time indexes of 2,4,...9998,10000
- the second df had time indexes 5002,5004...14998,15000
- concatenating the two files I obviously had duplicated time indexes, and preprocessing such as equilibrium detection failed. 

AMBER's pmemd/sander writes output information every ```ntpr``` timesteps, and so we are reading dHdL/u_nk values at this interval. The correct logic should be to multiply the index by the timestep, as it's already done, and then by this ```ntpr``` value. In this way, the time is correctly computed.

Interestingly enough, the FEData object already read the ```ntpr``` value from the AMBER output file, it's just then discarded and not saved as an attribute of the object, and never used. 

So in this fix  I just added ```ntpr``` to FEDdata ```__slots__```, added the read ```ntpr``` value as an attribute to the object, and then just changed the time computation to
```
file_datum.t0 + (frame_index + 1) * file_datum.dt * file_datum.ntpr
```


DrDomenicoMarson:
All the tests fail at the Codecov stage with an error 
```Error: Codecov: Failed to properly upload: ```
but I checked in the "Run tests" stage and everything's ok, I don't know what it's wrong!
Maybe the workflow I followed to make the pull request is wrong?

By the way, I was planning to make some other adjustments to the amber.py file, for example:
- it could be helpful to parse the file just one time if we need to extract both u_nk and dHdL values
- amber's output can be a little different in newer versions of pmemd.cuda, I can try to make the parser more "robust".

Let me know what you think about it!

**Comment by orbeckst** [comment]
Hi @DrDomenicoMarson , welcome to alchemlyb! Thank you for your PR.

Could you please raise an issue with the problem — we use the issue tracker to keep track of problems and there might be multiple PRs all for the same issue. You can just copy and paste your description from the PR. Then we can link the PR to the issue.

We like PRs to be focused on a single issue. This is cleaner and a lot easier to review. Please raise separate issues for other improvements to the AMBER parser. The smaller an issue and the PR the better!!!

Regarding the codecov error: That seems to be a Codecov issue. Maybe it is a temporary failure. I'll restart the CI and then we'll see.

**Comment by orbeckst** [comment]
The codecov error was temporary.

@xiki-tempula if you can spare some 👀 for a review....?

**Comment by xiki-tempula** [comment]
@DrDomenicoMarson Thank you very much for the PR. I wonder if you mind add a test to show that you have fixed the issue?
Is `ntpr` always present in the amber output?

**Review by orbeckst — CHANGES_REQUESTED** [review]
Please 
- raise an issue for the problem
- add a test
- add an entry to CHANGES referencing the issue
- add yourself to AUTHORS

and please address any points that @xiki-tempula is raising. Thanks! (sorry for the terse review, I don't have much time)

**Comment by DrDomenicoMarson** [comment]
Thank you everyone for your help!
@xiki-tempula I can assure you that ```ntpr``` is always present in an AMBER output file.
Unfortunately, I don't know how to properly add a test case for the issue, I'm completely new to "cooperative coding", I just write my python scripts usually.

I created the issue, now I have to leave for dinner, tomorrow morning I'll make another PR for this issue as you suggested!

**Comment by orbeckst** [comment]
@DrDomenicoMarson please do not open a new PR, just add to this one. All your new commits will show up here.

We can give you pointers on how to add tests.

**Comment by DrDomenicoMarson** [comment]
Hello everyone! Thanks again for your patience and your suggestions!

I hope I understood well how to operate pytest, I added the test for the new fix and it seems it is passing without problems. Also, I manually checked that the time information read now is the same reported in the AMBER output file (previously it wasn't).

I edited AUTHORS and CHANGES, I think everything should be fine now, right?

I started working on a more performing AMBER parser, next time I'll create the issue before and I'll create the PR from that,
with the tests already implemented (I hope)!

Please let me know if I have to do anything else!

File: src/alchemlyb/parsing/amber.py [hunk 1]
```diff
@@ -174,6 +174,7 @@ def __init__(self):
         self.t0 = -1.0
         self.dt = -1.0
         self.T = -1.0
+        self.ntpr = -1
```
xiki-tempula, line 177: Why do you set this to -1? 1 would make more sense.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -25,13 +25,16 @@ def invalid_file(request):
                           for filename in leg])
 def test_dHdl(filename,
```
xiki-tempula, line 26: It might be good to not modify the existing test but to add new tests.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -25,13 +25,16 @@ def invalid_file(request):
...
+              first_time=22.0,
+              last_time=1020.0):
     """Test that dHdl has the correct form when extracted from files."""
     dHdl = extract_dHdl(filename, T=300)
 
     assert dHdl.index.names == names
     assert dHdl.shape == shape
-
+    assert dHdl.index.values[0][0] == first_time
+    assert dHdl.index.values[-1][0] == last_time
```
xiki-tempula, line 37: You could just add a new test that checks for dHdl.index.values[0][0] and dHdl.index.values[-1][0]

Make sure to use `np.isclose` to check float point numbers.

xiki-tempula (changes requested):
Good work a couple of minor comments.

CHANGES
```diff
@@ -25,6 +25,7 @@ Enhancements
 
 Fixes
   - documented conda installation (available since 0.6.0) (#192)
+  - fixed how AMBER output parser reads time information (Issue #221, PR #219)
```
xiki-tempula, line 28: AMBER parser now use ntpr to reads time information.

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: I completely agree with you, but I set it to -1 as all the other defaults were settled to -1 (also the other defaults are meaningless in my opinion, but I didn't want to move too much away from the current implementation with this simple fix...) I think that for now, it's ok like this for coherence, but then I'll certainly change all the defaults when I'll refactor the parser in a future PR (I'm leaving for one week's vacation right now :-) )!

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: I've done so now, using ```isclose``` and also testing for the correct reading of u_nk values, as I had forgotten that previously!

CHANGES
DrDomenicoMarson: Yeah, this sounds much better, changed it :-)

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -23,15 +23,20 @@ def invalid_file(request):
...
-              names=('time', 'lambdas'),
-              shape=(500, 1)):
+def test_dHdl(filename, names=('time', 'lambdas'), shape=(500, 1)):
     """Test that dHdl has the correct form when extracted from files."""
     dHdl = extract_dHdl(filename, T=300)
-
     assert dHdl.index.names == names
     assert dHdl.shape == shape
 
+@pytest.mark.parametrize("filename",
```
xiki-tempula, line 32: For this PR, one test is enough, we don't want to increase the CI time with parametrize(.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -44,6 +49,14 @@ def test_u_nk(mbar_filename,
 
     assert u_nk.index.names == names
 
+@pytest.mark.parametrize("mbar_filename",
```
xiki-tempula, line 52: I think we don't need two tests for this? Assume that the files are not that different?

src/alchemlyb/parsing/amber.py
xiki-tempula: Thanks for the explanation. I do agree that consistency is good. However, I felt that the default value has to be something that is meaningful and can give reasonable results when a custom value is not provided. Writing a frame every -1 step will not give meaningful time. I do acknowledge that other default values are rubbish but for the new code, we want it to be something reasonable. What do you think @orbeckst ?

CHANGES
```diff
@@ -25,6 +25,7 @@ Enhancements
 
 Fixes
   - documented conda installation (available since 0.6.0) (#192)
+  - AMBER parser now use "ntpr" to reads time information. (Issue #221, PR #219)
```
xiki-tempula, line 28: Sorry might be my typo
```suggestion
  - AMBER parser now uses "ntpr" to read time information. (Issue #221, PR #219)
```

xiki-tempula (changes requested):
Great changes. A couple comments.

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: I think that for this PR we could be consistent with the old behavior (also having a dt of -1 doesn't make much sense). I'm planning to refactor the whole parser with a future PR, anyway!

DrDomenicoMarson: Also, these values are not provided by the user, they are read directly from the AMBER output file, and they are always present in every proper AMBER output file, so they shouldn't have a default value at all in my opinion.

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: The file is the same, but the MBAR energies and the dHdl values are in different places, and are read a little bit differently. Unfortunately right now the parser parses the file two times to read the different values, hence I wrote the two tests. In the future implementation I want to develop, both values will be read loading the file a single time, so the tests will be "merged".

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: This was my reasoning for which I tested for this fix in the old test. Testing for correct time-reading within the old test makes sense because the old test just checked if the dHdl/u_nk reader worked, checking for two parameters, I just added the time parameter to be reviewed, without increasing the time spent in "parameterize(". If you want I can revert to the old version which didn't slow down anything!

src/alchemlyb/parsing/amber.py
orbeckst: I'm ok with the rubbish -1 here in this PR.

In general I'd say `None` is a better default but let's leave these discussion for another PR. @DrDomenicoMarson if you plan to address this point, please open a new issue so that it's not forgotten.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -11,7 +11,7 @@
 from alchemtest.amber import load_invalidfiles
 from alchemtest.amber import load_bace_example
 from alchemtest.amber import load_bace_improper
-
+from numpy import isclose
```
orbeckst, line 14: Standard order of imports (standard lib, framework packages like numpy, specific packages like alchemtest, own modules) would have numpy imports before alchemtest... although I am not sure if we're consistent.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -23,15 +23,20 @@ def invalid_file(request):
                          [filename
                           for leg in load_simplesolvated()['data'].values()
                           for filename in leg])
-def test_dHdl(filename,
-              names=('time', 'lambdas'),
-              shape=(500, 1)):
+def test_dHdl(filename, names=('time', 'lambdas'), shape=(500, 1)):
```
orbeckst, line 26: Normally we don't touch lines for reformatting unless they raise PEP8 issues so that we reduce noise in the diff. 

But it's ok, leave it, now that you already changed it.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -44,6 +49,14 @@ def test_u_nk(mbar_filename,
 
     assert u_nk.index.names == names
 
+@pytest.mark.parametrize("mbar_filename",
+                         [mbar_filename
+                          for leg in load_bace_example()['data']['complex'].values()
+                          for mbar_filename in leg])
+def test_u_nk_time_reading(mbar_filename, first_time=22.0, last_time=1020.0):
+    u_nk = extract_dHdl(mbar_filename, T=300)
```
orbeckst, line 57: Do you really want to test `extract_dHdl()` here and not `extract_u_nk()`?

orbeckst (changes requested):
Overall this looks very good.  

I have one real question about why MBAR files are read with `extract_dHdl()` and not `extract_u_nk()`?  I realize that you only changed `extract_dHdl()` and I don't know much about AMBER FEP parsing so could you please explain the reasoning? Is there something that should be fixed in the AMBER MBAR reader as well?

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: You are right, it was a wild copy/paste :-)
I changed it in the last commit!

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: Ok, sorry I didn't know that!

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: Thanks, I fixed the order!

DrDomenicoMarson:
> 

I changed both functions because the "offending" line (the error) was the same for both parsers. AMBER output file is just one and contains both dHdl and u_nk information, so right now each function is reading the same file from the beginning, just extracting different information from it. So for now I had to change both functions to quickly fix the issue. 

The next version will be almost a rewrite of the parser, in which I plan to read the file just once if both u_nk and dHdl values are needed, refactoring out some common code of the two functions.

src/alchemlyb/tests/parsing/test_amber.py
xiki-tempula: So the incentive of the unit test is to make the debugging easier. When I change a piece of code, did it break the code? Is it like the code not running at all? Or it runs but didn't give the correct structure? Or it doesn't gives the correct numerical value?

The test `test_dHdl` tests if the function will run and it will not give out invalid results, so it covers all files but only check the basic structure of the output.

The test `test_dHdl_time_reading` will test if the correct time will be read.

So if `test_dHdl` passes, but `test_dHdl_time_reading` didn't, we know that the codes runs and the framework has been setup correctly but the time handling has been wrong.

If both of them failed, we know that the code didn't run at all.

Merging the two tests would make it harder to debug.

My suggestion is that instead of read all the files and check the time for each file, just run one file and check if the time is correct.

xiki-tempula (changes requested):
Good changes. If you could change test_u_nk_time_reading and test_dHdl_time_reading to only read one file, then this PR is good to go.

orbeckst (approved):
Thank you for addressing my points. Of course, please also address any outstanding issues that @xiki-tempula raised. I'll leave it to him to eventually finalize and squash-merge the PR.

orbeckst:
Why did you close this PR? 

Can we please finish this one before doing anything else? In particular, please don't add features to a PR that do not address issues raised in the comments as this makes it impossible for us to review it.

orbeckst:
@DrDomenicoMarson can you reopen the PR with the "Reopen and comment" button on GitHub?

DrDomenicoMarson:
I can't, or I'm not able to, I tried without success. 
I'll make a new fork, change only what's needed to address this issue (with all the comments you made about this PR), and make a new PR addressing only these changes, I think it will be easier that way.
The PR in alchemtest should be accepted before the new PR could pass the tests, anyway!

orbeckst:
stop, don't do anyhing yet

orbeckst:
This PR came from your master branch so we need to get your master branch back to commit 57aee2b288282ab7ee7baa1de0401bfa07f252a3 (Generally speaking, it's better to make branches in your fork for PRs (instead of working on master or main) and then create the PR from the branch.)

I think you should be able to save your other work (the one in PR #223) by making a branch *more-amber-parser-improvements* at [892142d](https://github.com/alchemistry/alchemlyb/pull/223/commits/892142d51d5d0476eb5d98a177cf671b3e8984e2)
```
git branch more-amber-parser-improvements  
```
Then switch to master
```
git switch master
```
and reset the master branch to 57aee2b288282ab7ee7baa1de0401bfa07f252a3
```
git reset --hard 57aee2b288282ab7ee7baa1de0401bfa07f252a3
```
Then push the updated master branch and this PR should update
```
git push -f 
```
to the previous stage.

The PR #223 will then be out of sync but that's ok. It can be closed. You'll create a new PR from your *more-amber-parser-improvements* branch.

Until this PR #219 is merged, you'll have to keep your local master branch clean and not add anything to it. Otherwise it will mess up this PR.

**If a step above does not work**, stop, send a message with the error output from git, and we try to help. Note that none of your work will be lost  because it's in branch *more-amber-parser-improvements*.

orbeckst:
See if you can reopen after doing the force-push.

DrDomenicoMarson:
The commands you gave worked, in my fork the master branch is what it should be, but I still can't reopen it, here is the error message: 

Screenshot 2022-09-13 at 00 11 49

I can create a new PR request now if you want! Is the PR in alchemtest ok?

orbeckst:
Your repo https://github.com/DrDomenicoMarson/alchemlyb/commits/master is not at 57aee2b288282ab7ee7baa1de0401bfa07f252a3
 but you have done other things (like merging the alchemlyb master as far as I can tell. 

At this stage I'm running out of time to reopen this particular PR. Please **open a new PR** that **only contains the patches in order to fix the ntpr issue**. Create it from a branch that is rooted in the alchemlyb tree and not in your master with additional modifications. Just ping @xiki-tempula and me when you have it up. We'll then re-review.

orbeckst:
>  Is the PR in alchemtest ok?

I haven't checked, it isn't necessary for this PR so it will have to wait a bit, sorry.

orbeckst:
(Or do you need additional test files for the ntpr issue?)

DrDomenicoMarson:
I made a new PR addressing just the ```ntpr``` issue, but yeah, the alchemtest PR I recently made is needed for this fix to pass tests properly.

Issue #226, opened 2022-09-13
AMBER: why are we reading AMBER output averages (and not using them)?


DrDomenicoMarson:
AMBER output can optionally report every ```ntave``` steps (if != 0) the average of dHdl and other parameters. 
Right now parsing dHdl we are collecting these values (here we are also collecting different components reported by AMBER), but we are not using any of this information at all in the code. Maybe it was a leftover from copying from ```alchemical-analysis```?. 

I think we could just skip the reading of the averages since we are collecting all the values (and the user usually performs better equilibration/subsampling analysis after reading all the data).

TO DO:
  - Skip the lines reporting the AVERAGES, instead of uselessly collecting them.

orbeckst:
I am happy with removing the data-reading that is not needed.

DrDomenicoMarson:
Great, I will address all these issues in some order, one at a time, to make the review process easier!

PR #230, opened 2022-09-14
Removed extraction of dHdl avg and components in AMBER parser

DrDomenicoMarson:
Addressing issue #226, I removed the useless extraction of dHdl averages (and dHdl components at the same time).

PS: I also made a really small adjustment to the ```skip_lines``` method of ```SectionParser```, as it stated it skipped a given number of "files" (????) while it should obviously be "lines" :-)

### DIFF
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -83,7 +83,7 @@ class SectionParser(object):
         self.lineno = 0
 
     def skip_lines(self, nlines):
-        """Skip a given number of files."""
+        """Skip a given number of lines."""
         lineno = 0
         for line in self:
             lineno += 1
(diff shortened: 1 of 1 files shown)


DrDomenicoMarson:
I see the code coverage went down a little bit, I don't think it is really a problem, but I don't understand why it decreased, as the new code should be executed during the tests, as it was before... 

Loosely related, I noticed there are some ```pragma: no cover``` comments, should I address them also in separate issues with new tests, right?

xiki-tempula:
> I see the code coverage went down a little bit, I don't think it is really a problem, but I don't understand why it decreased, as the new code should be executed during the tests, as it was before...

No worry. As long as the patch is good, it is fine. The project goes down as the number of the uncovered line remain unchanged while the total amount of lines has decreased.

> Loosely related, I noticed there are some pragma: no cover comments, should I address them also in separate issues with new tests, right?

Ideally yes, but adding new tests without changing the code base is usually fine.

xiki-tempula (approved):
LGTM. @orbeckst Shall I just merge this?

orbeckst:
Please squash-merge if you’re happy with it. Just synthesize all commit messages into something readable.

PR #231, opened 2022-09-16
ABFE workflow will not read dHdl file if no TI estimator is choosen

xiki-tempula:
If not explicitly set, AMBER will not generate the dHdl file. (I don't know if the u_nk will be written by default but I got a complain that workflow cannot be used to process their amber output, which doesn't have dhdl data). Currently, the ABFE workflow will attempt to read both dHdl and u_nk. This PR will make it such that if no TI estimator is chosen, the workflow will not attempt to read the dHdl.

I have also removed the ignore_warnings in the read file stage as maintaining it and having it to be clever enough to solve all the dumb user issues seems to be too much.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,9 @@ Changes
   - AutoMBAR accepts the `method` argument (PR #114).
   - Refactor the subsampling module to unify the behaviour of
     equilibrium_detection and statistical_inefficiency (PR #218).
+  - For ABFE workflow, dHdl will only be read when a TI estimator is chosen.
+    Similarly, u_nk will only be read when FEP estimators are chosen. (PR #231)
+
 
 Enhancements
   - Add u_nk2series and dhdl2series to convert u_nk and dHdl to series (PR
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,7 @@ Changes
     equilibrium_detection and statistical_inefficiency (PR #218).
   - For ABFE workflow, dHdl will only be read when a TI estimator is chosen.
     Similarly, u_nk will only be read when FEP estimators are chosen. (PR #231)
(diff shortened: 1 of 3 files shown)


xiki-tempula:
@orbeckst Do I need to update the changes? Technically nothing has been updated from the previous release so all this should be covered by the existing entry.

orbeckst:
We generally put all fixes in because in principle one can use the development version as we strive to never break it. It’s easier to decide that anything that’s a fix is recorded than to decide on a case by case basis.

xiki-tempula:
@orbeckst Thanks for the advice. I have made the changes and this PR is ready for review.

CHANGES
```diff
@@ -24,6 +24,10 @@ Changes
   - AutoMBAR accepts the `method` argument (PR #114).
   - Refactor the subsampling module to unify the behaviour of
     equilibrium_detection and statistical_inefficiency (PR #218).
+  - For ABFE workflow, dHdl will only be read when a TI estimator is chosen.
+    Similarly, u_nk will only be read when FEP estimators are chosen. (PR #231)
```
orbeckst, line 28: This seems more like a fix to me (reading the comment on the PR)

src/alchemlyb/workflows/abfe.py
```diff
@@ -48,8 +48,6 @@ class ABFE(WorkflowBase):
     outdirectory : str
         Directory in which the output files produced by this script will be
         stored. Default: os.path.curdir.
-    ignore_warnings : bool
-        Turn all errors into warnings.
```
orbeckst, line 52: I am supporting the "fail early and often approach" !

orbeckst (approved):
LGTM — I'd just move one of the CHANGELOG entries to fixes.

Please merge when you're happy with everything.

Issue #225, opened 2022-09-13
AMBER: temperature reading in parsers and tests. 


DrDomenicoMarson:
Temperature is required in input for both AMBER parsers (```extract_dHdl``` and ```extract_u_nk```), but also system temperature is already automatically read from the output file (reading ```temp0```). I understand making ```T``` optional create a problem with the decorator that adds ```T``` to the argument of the output from the parsers, so I think still requiring ```T``` as a parameter is fine.

Currently, if ```temp0``` is not found we log that 'Non-constant temperature MD not currently supported.' I don't think that this is right, as ```temp0``` is set even in non-constant temperature MD, so I think that the check as it is is meaningless. As a note, it could be possible to check if the simulation was performed not at constant T, but it's not as easy because this information is written at the end of the file in a format that it's not easy to catch with current SectionParser. This shouldn't be a big problem I think, as the user should know if the simulation he/she performed was done at constant T, and I don't know if other parsers make this check at all (maybe it was a left-over from copying from ```alchemical-analysis```?).

Still, we can benefit from reading ```temp0```, and issue a WARNING to the user if the ```T``` in input is different from what is read from the file.

Related to that, I noticed we set ```T=300``` in ```test_amber.py```, but the simulations are run at T=298.0 K, so this should be fixed as well.

RECAP of what can be done:
- still require ```T``` in input for the parser
- still read ```temp0``` from AMBER output
- remove the WARNING about non-constant T simulation
- add a WARNING if ```temp0``` != ```T```
- fix ```test_amber.py``` with T=298.0

Let me know if this is OK, I can do it as soon as PR #224 is successfully closed.

orbeckst:
Sounds all sensible to me. I'd be happy to review a PR that addresses this issue.

DrDomenicoMarson:
Great, I will address all these issues in some order, one at a time, to make the review process easier!

PR #232, opened 2022-09-18
AMBER parser now compare user-provided T with what's read from AMBER files

DrDomenicoMarson:
This PR addresses issue #225, now:

- we require ```T``` in input for the parser
- we still read ```temp0``` from AMBER output
- the WARNING about non-constant T simulation is now a WARNING (the check could also be removed)
- the parser raises a WARNING if ```temp0 != T``` (two new tests now assert that)
- now the tests in ```test_amber.py``` use the correct  T=298.0

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -36,6 +36,9 @@ Enhancements
 Fixes
   - documented conda installation (available since 0.6.0) (#192)
   - AMBER parsers now use 'ntpr' information to read time properly (#221)
+  - AMBER parsers now skip the reading of averages (#226)
+  - AMBER parsers now check if the T provided by the user is the same
+    at which the simulations were run (#225)
 
 
 07/22/2022 xiki-tempula, IAlibay, dotsdl, orbeckst, ptmerz
--- a/CHANGES
+++ b/CHANGES
@@ -35,10 +35,10 @@ Enhancements
 
 Fixes
   - documented conda installation (available since 0.6.0) (#192)
-  - AMBER parsers now use 'ntpr' information to read time properly (#221)
-  - AMBER parsers now skip the reading of averages (#226)
+  - AMBER parsers now use 'ntpr' information to read time properly (issue #221, PR #224)
+  - AMBER parsers now skip the reading of averages (issue #226, PR #230)
   - AMBER parsers now check if the T provided by the user is the same
-    at which the simulations were run (#225)
+    at which the simulations were run (issue #225, PR #232)
 
 
 07/22/2022 xiki-tempula, IAlibay, dotsdl, orbeckst, ptmerz
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -200,10 +200,8 @@ def file_validation(outfile):
                                           ['nstlim', 'dt'])
         T, = secp.extract_section('temperature regulation:', '^$',
                                   ['temp0'])
-        if not T:
-            logger.error('Non-constant temperature MD not '
-                         'currently supported.')
-            invalid = True
+        if not T:  # NOTE maybe we could remove this check completely
+            logger.warning('WARNING: no valid "temp0" record found in file')
         clambda, = secp.extract_section('^Free energy options:', '^$',
                                         ['clambda'], '^---')
         if clambda is None:
@@ -282,6 +280,12 @@ def extract_u_nk(outfile, T):
     file_datum = file_validation(outfile)
     if not file_validation(outfile):   # pragma: no cover
         return None
+
+    if not np.isclose(T, file_datum.T):
+        logger.warning(
+            'WARNING: the temperature read from the input file (%.2f K),'
+            ' is different from the temperature passed as parameter (%.2f K)' % (file_datum.T, T))
+
     if not file_datum.have_mbar:
         raise Exception('ERROR: No MBAR energies found! Cannot parse file.')
     with SectionParser(outfile) as secp:
@@ -342,6 +346,12 @@ def extract_dHdl(outfile, T):
     file_datum = file_validation(outfile)
     if not file_validation(outfile):
         return None
+
+    if not np.isclose(T, file_datum.T):
+        logger.warning(
+            'WARNING: the temperature read from the input file (%.2f K),'
+            ' is different from the temperature passed as parameter (%.2f K)' % (file_datum.T, T))
+
     finished = False
     comps = []
     with SectionParser(outfile) as secp:
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -282,9 +282,10 @@ def extract_u_nk(outfile, T):
         return None
 
     if not np.isclose(T, file_datum.T, atol=0.01):
-        logger.warning(
-            'WARNING: the temperature read from the input file (%.2f K),'
-            ' is different from the temperature passed as parameter (%.2f K)' % (file_datum.T, T))
+        msg = f'The temperature read from the input file ({file_datum.T:.2f} K)'
+        msg += f' is different from the temperature passed as parameter ({T:.2f} K)'
+        logger.error(msg)
+        raise ValueError(msg)
 
     if not file_datum.have_mbar:
         raise Exception('ERROR: No MBAR energies found! Cannot parse file.')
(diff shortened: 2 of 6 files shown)


src/alchemlyb/parsing/amber.py
```diff
@@ -282,6 +280,12 @@ def extract_u_nk(outfile, T):
     file_datum = file_validation(outfile)
     if not file_validation(outfile):   # pragma: no cover
         return None
+
+    if not np.isclose(T, file_datum.T):
+        logger.warning(
```
xiki-tempula, line 285: I would probably raise a Valueerror here, maybe cap the `atol` in `np.isclose` to 0.01.

src/alchemlyb/parsing/amber.py
```diff
@@ -342,6 +346,12 @@ def extract_dHdl(outfile, T):
     file_datum = file_validation(outfile)
     if not file_validation(outfile):
         return None
+
+    if not np.isclose(T, file_datum.T):
```
xiki-tempula, line 350: I would probably raise a Valueerror here, maybe cap the atol in np.isclose to 0.01.

xiki-tempula (changes requested):
Good changes. I think if the user-supplied T doesn't match the T in the file, a ValueError should be raised, as the user should know what they are doing and if the user says that the temperature is A but the real temperature is B, the user should change their input.

src/alchemlyb/parsing/amber.py
```diff
@@ -200,10 +200,8 @@ def file_validation(outfile):
                                           ['nstlim', 'dt'])
         T, = secp.extract_section('temperature regulation:', '^$',
                                   ['temp0'])
-        if not T:
-            logger.error('Non-constant temperature MD not '
-                         'currently supported.')
-            invalid = True
+        if not T:  # NOTE maybe we could remove this check completely
```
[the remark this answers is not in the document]
DrDomenicoMarson: As for T regulation in AMBER:
```temp0```: sets the target temperature at which te system is to be kept
```tempi```: sets the T according to wich the atoms' initial velocities are randomly assigned
no ```temp1``` exists :-)

In a production simulation, as the ones we are expecting to receive output from, ```tempi``` is not actually used, as these simulations are run from previous restarts files, from which velocities/T are automatically read.

If the variable ```nmropt>0``` and the namelist  ```&wt``` with ```type='TEMP0'``` is found, the target temperature can be modified during the simulation (e.g., we can raise the temperature from 0 to 300 K in the first 35000 steps and then keep it to 300 K for the remaining steps of the simulation)
```fortran
&wt type='TEMP0', istep1=0,istep2=35000,
value1=0.0, value2=300.0
```
This is how variable T simulations are done in AMBER, and I reported here the details as a future reference if we'll want to better address such cases. Reading such informations in the output file is a little more complicated with the current parser, so I think for now is better to assume the user knows he/she has to run in NVT/NPT ensamble.

DrDomenicoMarson:
> a ValueError should be raised

So do you want I change the WARNING to a proper ValueError that stops the parsing completely?

I agree this will be a more robust behavior, I just raised a warning to not break any existing script other users could be using right now!

We can wait for some input from @orbeckst regarding the rising of a WARNING/ERROR maybe?

CHANGES
```diff
@@ -36,6 +36,9 @@ Enhancements
 Fixes
   - documented conda installation (available since 0.6.0) (#192)
   - AMBER parsers now use 'ntpr' information to read time properly (#221)
```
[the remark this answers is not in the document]
DrDomenicoMarson: Yeah, changed it, this pattern makes lots of sense!

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: For now I added the ```atol=0.01``` parameter to ``isclose```, see the comment for WARNING/ValueError

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: For now I added the ```atol=0.01``` parameter to ```isclose```, see the comment for WARNING/ValueError

xiki-tempula:
Thanks for the comments and the commits. I think it is more like what kind of use cases that we want to handle. There are two user cases that which the user-supplied temperature could be different from the temperature in the file.

1. If the user supplied the wrong temperature and the user didn't know about this. 
2. The user supplies the temperature that is not the same as the one that is in the file and the user wishes to overwrite the temperature in the file.

My personal opinion is that I think for case 1, we raise an exception and ask the user to input the right temperature as we want the user to know what they are doing before going forward.
For case 2, I think we might not want the extract_ functions to handle this as the users who want to do this are expert users and they could modify the dataframe by themselves. For novice users, this might do more harm than good.

By the way, it might be good to not resolve the conversation and allow the reviewer to resolve the conversations. It might be my habit but I kind of use open conversation to track whether a comment has been resolved or not. Thanks for the contribution.

DrDomenicoMarson:
> Thanks for the comments and the commits. I think it is more like what kind of use cases that we want to handle. There are two user cases that which the user-supplied temperature could be different from the temperature in the file.
> 
> 1. If the user supplied the wrong temperature and the user didn't know about this.
> 2. The user supplies the temperature that is not the same as the one that is in the file and the user wishes to overwrite the temperature in the file.
> 
> My personal opinion is that I think for case 1, we raise an exception and ask the user to input the right temperature as we want the user to know what they are doing before going forward. For case 2, I think we might not want the extract_ functions to handle this as the users who want to do this are expert users and they could modify the dataframe by themselves. For novice users, this might do more harm than good.
> 
> It is concerning that the other parsers (like the gmx one) will just use the user-supplied temperature without checking if the user supplied temperature matches the temperature in the file or not.
> 
> By the way, it might be good to not resolve the conversation and allow the reviewer to resolve the conversations. It might be my habit but I kind of use open conversation to track whether a comment has been resolved or not. Thanks for the contributions.

Yeah, these are the two cases I was thinking about! So you say we raise an exception, so the beginner will be notified he has to check T, while the advanced user could just provide the same temperature as the input file and THEN modify the obtained dataframe according to its needs, right?

I'm sorry, I didn't know I had to leave the conversations unresolved :-)

xiki-tempula:
> So you say we raise an exception, so the beginner will be notified he has to check T, while the advanced user could just provide the same temperature as the input file and THEN modify the obtained dataframe according to its needs, right?

Yes.

> I'm sorry, I didn't know I had to leave the conversations unresolved :-)

No worry, I didn't know this as well until other people told me this.

DrDomenicoMarson:
Done, now using a wrong T raises the correct warning, and I updated the two tests to catch that.

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -53,23 +54,43 @@ def test_invalidfiles(invalid_file):
...
     assert_allclose(u_nk.index.values[0][0], first_time)
     assert_allclose(u_nk.index.values[-1][0], last_time)
 
 
+def test_wrong_T_should_raise_warning_in_extract_dHdl(single_dHdl, T=300.0):
+    """
+    Test if calling extract_dHdl with differnt T from what's
+    read from the AMBER file gives a warning
+    """
+    with pytest.raises(ValueError) as exc_info:
```
xiki-tempula, line 79: with pytest.raises(ValueError, match="is different from the temperature passed as parameter"):

xiki-tempula (changes requested):
Use match to match exception

DrDomenicoMarson:
Now a lots of tests are failing because they used the wrong T :-) I'm on it.... :-)

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: Thanks, changed that, will be in next commit!

DrDomenicoMarson:
While fixing other tests I catched these warnings in `alchemlyb/tests/test_ti_estimators.py`
```python
../../../usr/lib/python3/dist-packages/joblib/backports.py:7
  /usr/lib/python3/dist-packages/joblib/backports.py:7: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
    from distutils.version import LooseVersion

src/alchemlyb/tests/test_ti_estimators.py::test_TI_separate_dhdl_no_pertubed
  /home/domenico/sources/alchemlybDM/src/alchemlyb/tests/test_ti_estimators.py:160: FutureWarning: The 'inplace' keyword in DataFrame.set_index is deprecated and will be removed in a future version. Use `df = df.set_index(..., copy=False)` instead.
    dHdl.set_index('bound-lambda', append=True, inplace=True)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```
with pandas version 1.5.0rc0

DrDomenicoMarson:
> 

Now all tests should work, I changed T to match what was in the AMBER files, and changed E/err expected values accordingly!

xiki-tempula (approved):
LGTM. Thanks for addressing the comments and fixing the tests as well. Will merge once all the tests passes.

DrDomenicoMarson:
> LGTM. Thanks for addressing the comments and fixing the tests as well. Will merge once all the tests passes.

All passed now :-)

Issue #236, opened 2022-09-21
AMBER: why reinventing the wheel with any_none?


DrDomenicoMarson:
In the AMBER parser we have an amazing function

```python
def any_none(sequence):
    """Check if any element of a sequence is None."""

    for element in sequence:
        if element is None:
            return True

    return False
```

that is used just once on `mbar`, which is the return value of `extract_section`, which can return either `None` or a list of `int/float/None`
```python
                if any_none(mbar):
                    continue
```

Could we just replace this with

```python
if None in mbar:
    continue
```

and remove the custom-made function?

Moreover, we also have a test in place to check for this function, so we'll also save one test!

The only problem in using None in arises only if `mbar==None`, but the same problem would have arisen also with `any_none` as `NoneType` is not iterable, so we could either make the parser more robust here by checking also for the nature of `mbar`, or we could just make this change keeping the "risk factor" as it was before!

PR #237, opened 2022-09-21
replaced useless function "any_none()" with a check for "None in"

DrDomenicoMarson:
This PR removes the useless function `any_none()` which does the same work as an `if None in`, and the relative test is removed. 

This was suggested in issue #236.

I'm not introducing a check for `mbar==None`, maintaining the logic as before.


xiki-tempula (approved):
LGTM. Thanks. Will merge when the tests passes.

DrDomenicoMarson:
> LGTM. Thanks. Will merge when the tests passes.

That was a quick PR :-)

PR #239, opened 2022-09-23
add R_c (convergence fraction) convergence analysis

### LINKED ISSUE #104: add R_c (convergence fraction) convergence analysis
Add the analysis of converging fraction  R_c from the paper 

S. Fan, B. I. Iorga, and O. Beckstein. Prediction of octanol-water partition coefficients for the SAMPL6- log P molecules using molecular dynamics simulations with OPLS-AA, AMBER and CHARMM force fields. Journal of Computer-Aided Molecular Design, 34:543–560, 2020. https://doi.org/10.1007/s10822-019-00267-z 

cc @VOD555

xiki-tempula:
Fix #104 

I have reimplemented the R_c implemented by @VOD555.
I have also implemented the A_c based on my understanding of the paper.
 
I have done a speed check, for a 32 window simulation where each window has 10,000 data points.

```
from alchemlyb.convergence import A_c
import time

data = pd.Series(data=range(10000))
data.attrs['temperature'] = 310
data.attrs['energy_unit'] = 'kcal/mol'
a=time.time()
value = A_c([data, ] * 32)
b=time.time()
```

Takes 0.01 second, so I think the performance is satisfying.

Includes commits from 
- Fix dhdl2series and u_nk2series would not reattach the unit #248
- Change the plot_convergence so that it could take output from R_c #249
- Remove the 'dhdl' method in decorrelating the u_nk #250


xiki-tempula:
@orbeckst Do you mind review this PR, please?
Note that the coverage is not as high as we want due to

```
for i in range(out_length):
    if all(conv[i:]):
        return i / length
```

This loop checks of the last `i` points in `conv` are true. `conv` is true when the backward and the forward are all with a certain distance to the end point. Given that the last forward point must be the same as the last backward point as they both cover the whole trajectory, conv[-1] is always true and thus, the else part of this loop is never covered, which lowers the coverage.

orbeckst:
I might be able to look at it near the end of the week. Sorry, a bit busy.

@VOD555 , could you please review this PR? It's your method after all and you know it best. Although you cannot block or approve in alchemlyb, just state in your review what — if anything — needs to be done. I'll then use your review as a basis. Thank you!

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,135 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+    out_length = int(1 / precision)
+    g_forward = _cummean(array, out_length)
+    g_backward = _cummean(array[::-1], out_length)
+    length = len(g_forward)
+    # Final value
+    g = g_forward[-1]
+    conv = ((g_forward>g-diff) & (g_forward<g+diff)) & ((g_backward>g-diff)&(g_backward<g+diff))
+    for i in range(out_length):
+        if all(conv[i:]):
+            return i / length
```
orbeckst, line 223: Commenting on https://github.com/alchemistry/alchemlyb/pull/239/#issuecomment-1258090175 : How about adding
```python
else: # pragma: no cover
   raise AssertionError("Reached forbidden code path in convergence.R_c")
```

VOD555, line 223: It's better to also return `g_forward` and `g_backward` so that one can draw the convergence plot with these two series.

xiki-tempula:
@VOD555 Thanks for the review. I have returned the forward and backward as dataframe now.

orbeckst:
@xiki-tempula could you please resolve the conflicts with all the other recently merged PRs?

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+    whole trajectory is not equilibrated.
+
+    Parameters
+    ----------
+    series : pandas.Series
+        The input energy array.
+    precision : float
+        The precision of the output R_c.
+    tol : float
+        Tolerance of the convergence check in kcal/mol.
```
orbeckst, line 196: I think it makes more sense to use kT because that's what we use everywhere. 

If pushed to use an absolute unit then I'd rather go SI with kJ/mol... but for consistency I'd prefer kT.

Any opinions @VOD555 @xiki-tempula ?

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+            4  3.049769  3.037454            0.5
+            5  3.034078  3.040484            0.6
+            6  3.043274  3.032495            0.7
+            7  3.035460  3.036670            0.8
+            8  3.042032  3.046597            0.9
+            9  3.044149  3.044385            1.0
+
+    Note
+    ----
+    This function tries to compute R_c from equation 16 from
```
orbeckst, line 219: remove "tries" and use "computes"

docs/convergence.rst
```diff
@@ -33,6 +33,51 @@ Will give a plot looks like this
    A convergence plot of showing that the forward and backward has converged
    fully.
 
+Fraction Convergence
```
orbeckst, line 36: @VOD555 what is the name for the method?

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+
+    convergence = pd.DataFrame(
+        {'Forward': g_forward,
+         'Backward': g_backward,
+         'data_fraction': [i / length for i in range(1, length + 1)]})
+    convergence.attrs = series.attrs
+
+    # Final value
+    g = g_forward[-1]
+    conv = ((g_forward>g-tol) & (g_forward<g+tol)) & ((g_backward>g-tol)&(g_backward<g+tol))
```
orbeckst, line 241: Is this faster than abs?

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+
+    The output will the float R_c. R_c = 0 indicates that the system is well
+    equilibrated right from the beginning while R_c = 1 signifies that the
+    whole trajectory is not equilibrated.
+
+    Parameters
+    ----------
+    series : pandas.Series
+        The input energy array.
+    precision : float
```
orbeckst, line 193: Explain what the precision means (perhaps in a note). I don't think that this is part of the original formulation, or is it. (ping @VOD555, too )

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+
+    The input is a list of pandas.Series generated by
+    :func:`~alchemlyb.preprocessing.subsampling.decorrelate_u_nk` or
+    :func:`~alchemlyb.preprocessing.subsampling.decorrelate_dhdl`.
+
+    The output will the float A_c. A_c is a number between 0 and 1 that can be
+    interpreted as the ratio of the total equilibrated simulation time to the
+    whole simulation time for a full set of simulations. A_c = 1 means that all
+    simulation time frames in all windows can be considered equilibrated, while
+    A_c = 0 indicates that nothing is equilibrated.
```
orbeckst, line 265: use math mode for $A_c$ and equations like $A_c = 1$

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+    return result
+
+def R_c(series, precision=0.01, tol=2):
+    '''Generate the convergence criteria R_c for a single simulation.
+
+    The input will be pandas.Series generated by
+    :func:`~alchemlyb.preprocessing.subsampling.decorrelate_u_nk` or
+    :func:`~alchemlyb.preprocessing.subsampling.decorrelate_dhdl`.
+
+    The output will the float R_c. R_c = 0 indicates that the system is well
```
orbeckst, line 185: math mode $R_c$

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+        Tolerance of the convergence check in kcal/mol.
+
+    Returns
+    -------
+    float
+        The area under curve for convergence time fraction.
+
+    Note
+    ----
+    This function tries to compute R_c from equation 18 from
```
orbeckst, line 283: This function computes $R_c$ ...

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+    A_c = 0 indicates that nothing is equilibrated.
+
+    Parameters
+    ----------
+    series_list : list
+        A list of pandas.Series energy array.
+    precision : float
+        The precision of the output R_c.
+    diff : float
+        Tolerance of the convergence check in kcal/mol.
```
orbeckst, line 274: in kT ?

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,166 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+    interpreted as the ratio of the total equilibrated simulation time to the
+    whole simulation time for a full set of simulations. A_c = 1 means that all
+    simulation time frames in all windows can be considered equilibrated, while
+    A_c = 0 indicates that nothing is equilibrated.
+
+    Parameters
+    ----------
+    series_list : list
+        A list of pandas.Series energy array.
+    precision : float
```
orbeckst, line 271: possibly refer to R_c() for detailed explanation

src/alchemlyb/convergence/convergence.py
orbeckst: Do you really mean R_c or should it be A_c.

I'd also put a SeeAlso to R_c and state that the R_c are calculated for the whole set using `R_c()`.

orbeckst (changes requested):
Overall this looks very good. I have some questions and would also like @VOD555 's input.

For consistency I'd keep input in $k_B T$ but if you have good reasons for absolute energies please say so.

docs/convergence.rst
VOD555: It originated from Yang W, Bitetti-Putzer R, Karplus M's paper, "Free energy simulations: Use of reverse cumulative averaging to determine the equilibrated region and the time required for convergence". Maybe we can call our method forward & reverse cumulative averaging.

src/alchemlyb/convergence/convergence.py
VOD555: It's not in the original formulation.

src/alchemlyb/convergence/convergence.py
VOD555: I prefer KT.

docs/convergence.rst
xiki-tempula: We have already had one forward_backward_convergence
https://alchemlyb.readthedocs.io/en/latest/convergence/alchemlyb.convergence.convergence.html#alchemlyb.convergence.forward_backward_convergence

Maybe we can call this cumulative_averaging?

orbeckst: How about "reverse cumulative averaging" — quite similar to the original paper?

We should then also mention the Yang paper. @VOD555 can you please make a suggestion on the PR how to change the docs?

VOD555: Then call it "reverse cumulative averaging". I think we should mention that the ` forward_backward_convergence` is done on all samplings for intermediate states, and the "reverse cumulative averaging" here is for a single trajectory.

xiki-tempula: I'm not sure if `reverse cumulative averaging` is good as `forward cumulative averaging` is done as well. I think the `forward & reverse cumulative averaging` is good for description but is too long for the name.

orbeckst: How about `fwdrev_cum_avg()` or `FR_cumulative_average()` for a function name? The method name itself could be 
*Forward/reverse cumulative averaging* as it doesn't really matter how many letters it takes up. It is then similar to the original Yang paper but also includes the "forward" component, which is important for the convergence condition.

xiki-tempula: I like the name of fwdrev_cum_avg. For the function, we have `R_c` and `A_c`, if we rename the function `R_c` to `fwdrev_cum_avg`, what should we do with A_c?

orbeckst: I re-read the documentation and looked at things in context. It reminded me that the *fractional* part is an important detail that makes everything comparable. I'd therefore suggest a section title **Fractional equilibration time** — that's pretty much what we called $R_c$ in the paper.

I still like `R_c` but I also understand that the function returns _both_ $R_c$ and the data. I'd call the current `R_c()` function  `fwdrev_cumavg_Rc()` (numpy has `cumsum()` so `cumavg` should look familiar, `fwdrev` nods to the original method, and `Rc` connects it to the fractional equilibration time) — a bit clunky but at least gives an idea that one gets multiple things back.

I'd keep `A_c()` as it makes it easy to relate it to the math and the paper.

Sorry for the back and forth... that's now my opinion and I am going to stand by it. Let me know what you think.

p.s.: The image should say "energy/kT" instead of $\Delta G/kT$.

xiki-tempula: I think this is reasonable, will address the comments this weekend.

docs/convergence.rst
```diff
@@ -33,6 +33,51 @@ Will give a plot looks like this
...
+    >>> decorrelated = decorrelate_dhdl(dhdl, remove_burnin=True)
+    >>> value, running_average = R_c(dhdl2series(decorrelated), tol=2)
+    >>> print(value)
+    0.02
+    >>> plot_convergence(running_average, final_error=2, units='kcal/mol')
+
+
+Will give a plot like this.
+
+.. figure:: images/R_c.png
```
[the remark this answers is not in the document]
xiki-tempula: I had a quick think of it. I think this is not necessarily true, the y-axis could also be the u_nk.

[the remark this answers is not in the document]
orbeckst: I suppose so. I don't think we tried it for u_nk. But for the example it's a dH/dl.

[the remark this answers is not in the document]
xiki-tempula: I will change this to
>>> ax = plot_convergence(running_average, final_error=2, units='kcal/mol')
>>> ax.set_ylabel("$\partial H/\partial\lambda$ (in $kT$)")

And then put the right figure here.

xiki-tempula:
@orbeckst Thanks very much for the review. Sorry for the delay. I have addressed all the issues.

CHANGES
```diff
@@ -39,6 +39,7 @@ Enhancements
   - Add remove_burnin keyword to decorrelate_u_nk and decorrelate_dhdl (PR #218).
   - Add a base class for workflows (PR #188).
   - Add the ABFE workflow (PR #114).
+  - Add R_c and A_c for convergence analysis. (PR #239)
```
orbeckst, line 42:
```suggestion
  - Add R_c and A_c for "fractional equilibration time" convergence analysis 
    (issue #104, PR #239)
```

docs/convergence.rst
```diff
@@ -33,6 +33,52 @@ Will give a plot looks like this
...
+    >>> from alchemlyb.parsing.gmx import extract_dHdl
+    >>> from alchemlyb.preprocessing import decorrelate_dhdl, dhdl2series
+    >>> from alchemlyb.convergence import fwdrev_cumavg_Rc
+    >>> from alchemlyb.visualisation import plot_convergence
+
+    >>> file = load_ABFE().data['ligand'][0]
+    >>> dhdl = extract_dHdl(file, T=300)
+    >>> decorrelated = decorrelate_dhdl(dhdl, remove_burnin=True)
+    >>> value, running_average = fwdrev_cumavg_Rc(dhdl2series(decorrelated), tol=2)
+    >>> print(value)
```
orbeckst, line 57: Make clearer what the value is:
```suggestion
    >>> R_c, running_average = fwdrev_cumavg_Rc(dhdl2series(decorrelated), tol=2)
    >>> print(R_c)
```

docs/convergence.rst
```diff
@@ -33,6 +33,52 @@ Will give a plot looks like this
...
+    0.04
+    >>> ax = plot_convergence(running_average, final_error=2, units='kcal/mol')
```
orbeckst, line 59: I think the units are wrong here, especially as the plot says kT
```suggestion
    >>> ax = plot_convergence(running_average, final_error=2, units='kT')
```

orbeckst: Is the `final_error` here the same as the value for `tol` ? If so, comment or use a variable.

docs/convergence.rst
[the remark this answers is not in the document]
orbeckst: The current figure has "energy/kT" on the y-axis. I think $\partial H/\partial\lambda$ (in kT)" is clearer.

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,171 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+            7  3.035460  3.036670            0.8
+            8  3.042032  3.046597            0.9
+            9  3.044149  3.044385            1.0
+
+    Note
+    ----
+    This function computes :math:`R_c` from equation 16 from
+    https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8397498/#FD16. The code is
+    modified based on Shujie Fan's (@VOD555) work.
+
```
orbeckst, line 224:
```suggestion
   
   Please cite [Fan2021]_ when using this function.
   
```
Can you check that the reference resolves?

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,171 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+        Tolerance of the convergence check in :math:`kT`.
+
+    Returns
+    -------
+    float
+        The area under curve for convergence time fraction.
+
+    Note
+    ----
+    This function tries to compute :math:`A_c` from equation 18 from
```
orbeckst, line 287:
```suggestion
    This function computes :math:`A_c` from equation 18 from
```

src/alchemlyb/convergence/convergence.py
```diff
@@ -137,3 +138,171 @@ def forward_backward_convergence(df_list, estimator='MBAR', num=10, **kwargs):
...
+    https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8397498/#FD18. See also
+    :func:`~alchemlyb.convergence.fwdrev_cumavg_Rc`.
+
```
orbeckst, line 290:
```suggestion

    Please cite [Fan2021]_ when using this function.

```

orbeckst (changes requested):
Just minor doc things, otherwise looks good.

@VOD555 do you want to take a last look?

xiki-tempula:
@orbeckst Thanks for the review. I have checked that the reference could be resolved.

docs/convergence.rst
```diff
@@ -45,4 +91,13 @@ The currently available connvergence functions:
     :toctree: convergence
 
     convergence
```
orbeckst, line 93:
```suggestion
    forward_backward_convergence
```

orbeckst (approved):
Thanks for addressing all my nitpicks, much appreciated.

orbeckst:
I made one change to the docs (wasn't in your edits) and I am now waiting for CI.

orbeckst:
Thank you @xiki-tempula , a very welcome contribution!

orbeckst:
I added @VOD555 as a "co-authored" by to the squash-merge because you said that you used his code as an initial basis and because he participated in the review. I put an extra note in the commit message that this version has your performance improvement included.

PR #243, opened 2022-09-27
Add extract description in docs

DrDomenicoMarson:
Just a couple of things that I forgot about:
- added the description of `extract` to the main `parsers` page in the documentation
- the test in test_namd.py now properly tests if 'dHdl' is not set when calling `extract`


xiki-tempula (approved):
LGTM

PR #248, opened 2022-10-05
Fix dhdl2series and u_nk2series would not reattach the unit

xiki-tempula:
Fix dhdl2series and u_nk2series would not reattach the unit


orbeckst (approved):
lgtm, thanks!

Issue #247, opened 2022-10-05
Default the unit of the plotting function to None


xiki-tempula:
As discussed in https://github.com/alchemistry/alchemlyb/issues/244#issuecomment-1264356239
Currently plot_dF_state, plot_ti_dhdl and plot_convergence all have `kT` as default unit. 
This kind of creates extra confusions, as the input estimator or dataframe has already had unit and the plotting function will force them to be `kT`.

I purpose that 

- The unit is default to None and the plotting function will just use the unit already in the estimator or dataframe. If the unit is not None, we would then do the unit conversion.
- Given that we are inputing dataframe to plot_convergence, I'm also thinking of deprecating the unitless array input to plot_convergence to make life easier.

This is partially linked to #246 as currently, the user could change the `delta_f_` and `d_delta_f_`, and then different part of the estimator could have different units. But if the #246 is fixed, we could be sure that the estimator has one unit.

What do you think @orbeckst

DrDomenicoMarson:
I really like this new approach! (I now I was never asked  for an opinion, but I really like it :-) )!

orbeckst:
You (and anyone else using or developing alchemyb) are _always_ welcome to offer opinions and review PRs!!!

Please do hit the "review" button on any PR that you're looking at. Constructive feedback or general agreement is always helpful.

orbeckst:
@xiki-tempula , I like the idea, too.

Sorry for my delayed response, I just missed it. (Thanks to @DrDomenicoMarson comment it appeared again on my radar. See @DrDomenicoMarson , your comment alone is moving us forward ;-) ).

DrDomenicoMarson:
Great! I can't promise that I'll have time in the next few days to work on this, but I'll be happy to contribute if I manage to!

PR #249, opened 2022-10-05
Make the error optional for plot_convergence

xiki-tempula:
Currently the plot_convergence demand the input dataframe to have forward_error and backward_error. This PR makes these two input optional. So when these two columns are not present in the input dataframe, no errorbar will be plotted.

I have not touched the array input as we will deprecate them in #247.

A final_error keyword has been added to deal with the case where error is not provided and the user wish to plot the error of the final estimate.

Fix the issue where when too many data points are included, the x-axis will be too crowded.

### DIFF
--- a/src/alchemlyb/visualisation/convergence.py
+++ b/src/alchemlyb/visualisation/convergence.py
@@ -115,7 +115,7 @@ def plot_convergence(*data, units='kT', final_error=None, ax=None):
 
     xticks_spacing = len(r_ts) // 10 or 1
     xticks = r_ts[::xticks_spacing]
-    plt.xticks(xticks, fontsize=10)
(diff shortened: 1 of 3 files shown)


src/alchemlyb/visualisation/convergence.py
```diff
@@ -23,6 +23,8 @@ def plot_convergence(*data, units='kT', ax=None):
     shape and can be used as input for the
     :func:`matplotlib.pyplot.errorbar`.
 
+    final_error is a float decides the width of the bar around the final value.
```
orbeckst, line 26:
```suggestion
   `final_error` is the error of the final value and is shown as the error band around the 
   final value. It can be provided in case an estimate is available that is more appropriate
   than the default, which is the error of the last value in `backward`.
```

orbeckst (approved):
lgtm

PR #252, opened 2022-10-16
delta_f_, d_delta_f_, states_ are view of the original object in estimator.

### LINKED ISSUE #246: User shouldn't be able to change the estimator attribute
As discussed in https://github.com/alchemistry/alchemlyb/issues/244#issuecomment-1264352791
Currently we allow the user to change the attribute in the estimator object. This would mean that the user could do
```python
ti = TI().fit(df)  # df is in kcal/mol
# now  ti.delta_f_, ti.d_delta_f_, and ti.dhdl are all in kcal/mol
ti.delta_f_ = to_kT(ti.delta_f_)
# now the user has an estimator with delta_f_ values in kT and the d_delta_f_ in kcal/mol
```
Example given by @DrDomenicoMarson 
Which could cause confusion, I'm thinking of make all these delta_f_ d_delta_f_ into a property which only has the get method and thus prevent the user from editing them.

Do you think this could cause issues? @orbeckst

xiki-tempula:
Fix #246


src/alchemlyb/estimators/__init__.py
```diff
@@ -1,3 +1,21 @@
+class EstimatorMixOut():
+    '''This class creates view for the d_delta_f_, delta_f_, states_ for the
+    estimator class to consume.'''
+    _d_delta_f_ = None
+    _delta_f_ = None
+    _states_ = None
```
orbeckst, line 6: Using the class-level attributes here is a bit tricky: they act as "read" defaults. They will only provide a value of `None` when `self._d_delta_f_` has not been set yet. As soon as `self._d_delta_f_ = value` has been assigned, an instance level attribute is created with `value`, which shadows the class level ones.

It wasn't immediately obvious to me that this would work but I think it does. (I was worried that attributes would be shared between instances but as long as nobody changes the base class attributes themselves it should be fine.)

src/alchemlyb/estimators/__init__.py
```diff
@@ -1,3 +1,21 @@
+class EstimatorMixOut():
```
orbeckst, line 1: move to a separate file (base.py) to avoid circular imports

src/alchemlyb/estimators/bar_.py
```diff
@@ -2,11 +2,11 @@
 import pandas as pd
 
 from sklearn.base import BaseEstimator
-
 from pymbar import BAR as BAR_
 
+from . import EstimatorMixOut
```
orbeckst, line 7: circular import — avoid (see above: move EstimatorMixOut to separate file)

src/alchemlyb/estimators/mbar_.py
```diff
@@ -3,10 +3,11 @@
 import logging
 
 from sklearn.base import BaseEstimator
-
 import pymbar
 
-class MBAR(BaseEstimator):
+from . import EstimatorMixOut
```
orbeckst, line 8:
```suggestion
from .base import _EstimatorMixOut
```

src/alchemlyb/estimators/bar_.py
orbeckst:
```suggestion
from .base import _EstimatorMixOut
```

src/alchemlyb/estimators/ti_.py
```diff
@@ -3,8 +3,9 @@
 
 from sklearn.base import BaseEstimator
 
+from . import EstimatorMixOut
```
orbeckst, line 6:
```suggestion
from .base import _EstimatorMixOut
```

src/alchemlyb/estimators/__init__.py
orbeckst, line 1: Name `_EstimatorMixOut` so that we can keep it internal and don't have to worry about documenting it.

orbeckst (changes requested):
Avoid circular imports, and I'd make the new mixin private, otherwise looks good.

src/alchemlyb/estimators/base.py
```diff
@@ -0,0 +1,17 @@
...
+    def d_delta_f_(self):
+        return self._d_delta_f_
+
+    @property
+    def delta_f_(self):
+        return self._delta_f_
+
+    @property
+    def states_(self):
+        return self._states_
```
orbeckst, line 17: add trailing \n
```suggestion
        return self._states_
        
```

orbeckst (approved):
Thanks for addressing my comments — all lgtm!

orbeckst:
@xiki-tempula can you please look into why coverage in abfe.py dropped?

orbeckst:
This what codecov says https://app.codecov.io/gh/alchemistry/alchemlyb/pull/252
image

xiki-tempula:
@orbeckst I suspect this is a codecov glitch. Locally, pycharm told me that this line has been covered.

xiki-tempula:
It is really weird. Locally, the ABFE workflow is 100% coverage.
Screenshot 2022-10-20 at 17 51 25

orbeckst:
I know, I did the same thing with coverage. I will merge as is. I just had to rerun one runner.

Issue #238, opened 2022-09-21
AMBER: are we too permissive in reading the files? (and also we are doing a double-reading?


DrDomenicoMarson:
Here I want to tackle an issue that may change the behavior of some existing scripts, but I think for the better.

Currently, in the AMBER parser, we have a function called `file_validation` (the name is pretty misleading because indeed it validates the file, but more importantly it returns either `False` or an object representation of the input AMBER file, on which then the other functions operate).
This function, which should return a "file parsed object `FEData`", whenever it encounters some problem reading the file, just prints a descriptive WARNING and switches an internal variable `invalid` to `True`, then continues reading the file as it nothing happened, only to return `False` if `invalid` was `True` instead of returning the parsed (and invalid) object.
This function is called when extracting dHdl and u_nk, and here is funny, because in both cases we do:

```python
file_datum = file_validation(outfile)
if not file_validation(outfile):  # pragma: no cover
    return None
```
which in itself is redundant, as it was sufficient to check `if not file_datum: return None` without parsing the same file twice.

The flow right now then is that we read and parse the AMBER output file to create the `FEData` object, if the file it's ok we continue parsing the AMBER out to get the data we need, otherwise, both `extract_*` functions return `None`.

I'm proposing some changes here (which have not to be implemented altogether)

1. remove the redundant `file_Validation` call (this I think must be done 😆 )
2. in `file_validation`, as soon as problems that now make the function return `False`, instead of switching `invalid=True`, raise a proper Exception (maintaining the current output warning). In this way, we won't continue parsing a file we already decided is invalid... after that we have three scenarios:
   2a.  Maintain the current behavior, we catch the exception when we use `file_validation`, like
          `try: file_datum = file_validation(outfile) ; except ExceptionWeWillRise: return None`
   2b. We don't catch the exception, so instead of returning `None` when an AMBER output file was not read we let the `file_validation` interrupt the flow of the program and let the user be aware that a file was not parsed.
   2c. An hybrid version of 3a and 3b, we catch the exception from `file_validation` inside the `extract_*` functions, and raise here a more informative exception, maybe printing both a general warning here and also the warning from the exception risen inside `file_validation`
3. change the name of the function `file_validation` to something more coherent to its behavior, I'm proposing `file_initialiation`, but I'm open to any suggestion.

I would certainly want to do 1., probably want to do 3. and I propend towards 2b. as it will let the users be aware that a file is not read, letting them be able to just catch the exception if they want to continue parsing more files.

[...]

PR #253, opened 2022-10-16
Use new testfiles and increase coverage to 100% for amber parser

DrDomenicoMarson:
I removed the dependency to load_invalidfiles and replaced it with the new load_testfiles, now I removed all  # pragma and the coverage for the parser should be 100%

### DIFF
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -192,10 +192,10 @@ def file_validation(outfile):
             invalid = True
 
         mbar_ndata = 0
-        have_mbar, mbar_ndata = secp.extract_section('^FEP MBAR options:',
+        have_mbar, mbar_ndata, mbar_states = secp.extract_section('^FEP MBAR options:',
                                                       '^$',
                                                       ['ifmbar',
-                                                        'bar_intervall'],
+                                                        'bar_intervall', "mbar_states"],
                                                       '^---')
         if have_mbar:
             mbar_ndata = int(nstlim / mbar_ndata)
@@ -210,6 +210,11 @@ def file_validation(outfile):
                 have_mbar = False
             else:
                 mbar_nlambda = len(mbar_lambdas)
+                if mbar_nlambda != mbar_states:
+                    logger.warning(
+                        "The number of lambda windows read (%s)"
+                        " is differemt from what expected (%d)",
+                        ','.join(mbar_lambdas), mbar_states)
                 mbar_lambda_idx = mbar_lambdas.index(clambda_str)
                 file_datum.mbar_lambda_idx = mbar_lambda_idx
 
@@ -418,7 +423,7 @@ def _process_mbar_lambdas(secp):
             if 'total' in line:
                 data = line.split()
                 mbar_lambdas.extend(data[2:])
-            else: # pragma: no cover
+            else:
                 mbar_lambdas.extend(line.split())
 
     return mbar_lambdas
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -424,6 +424,8 @@ def _process_mbar_lambdas(secp):
                 data = line.split()
                 mbar_lambdas.extend(data[2:])
             else:
+                # AMBER splits the MBAR states in more than one line 
+                # if there are many of them
                 mbar_lambdas.extend(line.split())
 
     return mbar_lambdas
--- a/src/alchemlyb/parsing/amber.py
+++ b/src/alchemlyb/parsing/amber.py
@@ -65,8 +65,10 @@ class SectionParser():
         self.filename = filename
         try:
             self.fileh = anyopen(self.filename, 'r')
-        except Exception as ex:  # pragma: no cover
-            logger.exception("Cannot open file %s", filename)
+        except Exception as ex:
+            msg = f"Cannot open file {filename}"
+            logger.exception(msg)
+            raise FileNotFoundError(msg) from ex
         self.lineno = 0
 
     def skip_lines(self, nlines):
@@ -288,7 +290,10 @@ def extract(outfile, T):
                 mbar = secp.extract_section('^MBAR', '^ ---', file_datum.mbar_lambdas,
                                             extra=line)
                 
-                if None in mbar: # pragma: no cover
+                if None in mbar:
+                    msg = "WARNING, something strange parsing the following MBAR section."
+                    msg += "\nMaybe the mbar_lambda values are incorrect?"
+                    logger.warning(f"{msg}\n{mbar}")
                     continue
                 
                 reference_energy = mbar[file_datum.mbar_lambda_idx]
@@ -305,7 +310,7 @@ def extract(outfile, T):
             logger.warning('%i MBAR energ%s > 0.0 kcal/mol',
                            high_E_cnt, 'ies are' if high_E_cnt > 1 else 'y is')
 
-    if not finished: # pragma: no cover
+    if not finished:
         logger.warning('WARNING: file %s is a prematurely terminated run' % outfile)
 
     if file_datum.have_mbar:
@@ -323,7 +328,7 @@ def extract(outfile, T):
         logger.info('WARNING: No MBAR energies found! "u_nk" entry will be None')
         mbar_df = None
 
-    if not nensec: # pragma: no cover
+    if not nensec:
         logger.warning('WARNING: File %s does not contain any dV/dl data' % outfile)
         dHdl_df = None
     else:
--- a/src/alchemlyb/tests/parsing/test_amber.py
+++ b/src/alchemlyb/tests/parsing/test_amber.py
@@ -1,29 +1,105 @@
 """Amber parser tests.
 
 """
+import logging
 import pytest
 from numpy.testing import assert_allclose
 
 from alchemlyb.parsing.amber import extract_dHdl
 from alchemlyb.parsing.amber import extract_u_nk
-from alchemlyb.parsing.amber import file_validation
 from alchemlyb.parsing.amber import extract
 from alchemtest.amber import load_simplesolvated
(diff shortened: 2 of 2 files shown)


src/alchemlyb/parsing/amber.py
```diff
@@ -65,8 +65,10 @@ def __init__(self, filename):
         self.filename = filename
         try:
             self.fileh = anyopen(self.filename, 'r')
-        except Exception as ex:  # pragma: no cover
-            logger.exception("Cannot open file %s", filename)
+        except Exception as ex:
+            msg = f"Cannot open file {filename}"
+            logger.exception(msg)
+            raise FileNotFoundError(msg) from ex
```
orbeckst, line 71: Why on earth did we not even raise the exception there... ? Thanks for raising.

However, it's not clear that this is a FileNotFoundError. Let's not pretend that we know what went wrong. Just reraise the original exception
```suggestion
            raise
```

src/alchemlyb/parsing/amber.py
```diff
@@ -65,8 +65,10 @@ def __init__(self, filename):
         self.filename = filename
         try:
             self.fileh = anyopen(self.filename, 'r')
-        except Exception as ex:  # pragma: no cover
-            logger.exception("Cannot open file %s", filename)
+        except Exception as ex:
+            msg = f"Cannot open file {filename}"
+            logger.exception(msg)
```
orbeckst, line 70: Keep the original logger line as we don't need to repeat the message in the exception.
```suggestion
            logger.exception("Cannot open file %s", filename)
```

src/alchemlyb/parsing/amber.py
```diff
@@ -190,10 +192,10 @@ def file_validation(outfile):
             invalid = True
 
         mbar_ndata = 0
-        have_mbar, mbar_ndata = secp.extract_section('^FEP MBAR options:',
+        have_mbar, mbar_ndata, mbar_states = secp.extract_section('^FEP MBAR options:',
                                                       '^$',
                                                       ['ifmbar',
-                                                        'bar_intervall'],
+                                                        'bar_intervall', "mbar_states"],
```
orbeckst, line 198: What is this change doing? Is it changing the functionality? If so, it needs an CHANGELOG and versionchanged entry.

Also, just for optical consistency use the same quote characters and add it on a new line.

orbeckst: Does parsing this section affect reading performance?

src/alchemlyb/parsing/amber.py
```diff
@@ -208,6 +210,11 @@ def file_validation(outfile):
                 have_mbar = False
             else:
                 mbar_nlambda = len(mbar_lambdas)
+                if mbar_nlambda != mbar_states:
+                    logger.warning(
+                        "The number of lambda windows read (%s)"
+                        " is differemt from what expected (%d)",
+                        ','.join(mbar_lambdas), mbar_states)
```
orbeckst, line 217: Shouldn't this be an exception?

src/alchemlyb/parsing/amber.py
```diff
@@ -288,7 +295,10 @@ def extract(outfile, T):
                 mbar = secp.extract_section('^MBAR', '^ ---', file_datum.mbar_lambdas,
                                             extra=line)
                 
-                if None in mbar: # pragma: no cover
+                if None in mbar:
+                    msg = "WARNING, something strange parsing the following MBAR section."
+                    msg += "\nMaybe the mbar_lambda values are incorrect?"
+                    logger.warning(f"{msg}\n{mbar}")
```
orbeckst, line 301: Makes me shudder that we just continue here instead of failing. We obviously had this code here before but it makes me uncomfortable that we allow continuing and potentially creating wrong results. Or would this be considered a normal situation when working with AMBER files??

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -1,30 +1,123 @@
...
+
+
+def test_file_not_found():
+    with pytest.raises(FileNotFoundError):
+        _ = extract("not_a_valid_file_name", T=298.0)
+
+
+def test_no_dHdl_data_points(caplog, testfiles):
+    """Test if we deal with a file without dHdl data points"""
+    filename=testfiles["no_dHdl_data_points"][0]
```
orbeckst, line 35: pep8: spaces around assignment op

orbeckst: Or make it a kwarg
```python
def test_no_dHdl_data_points(caplog, testfiles,  filename=testfiles["no_dHdl_data_points"][0]):
```

orbeckst (changes requested):
Thank you for updating the tests. Please see comments inline. Thanks!

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: I don't think this will affect reading performance, but I hadn't tested it. We were already parsing the same line, but instead of reading 2 items now we read 3, not a big change. This is needed to check if there is consistency in the number of MBAR states read and expected by AMBER, so it's a value never exposed to the user and it doesn't change the overall functionality!

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: I think so, I just wrote a warning because with the previous version we were not rising exceptions when this section wasn't parsed correctly, but I agree we should.

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: As far as I know, this is not considered a normal situation with AMBER files, I kept things as they were, but I agree a proper exception is better suited here.

src/alchemlyb/tests/parsing/test_amber.py
DrDomenicoMarson: I added the spaces!

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: Done!

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: Done!

src/alchemlyb/parsing/amber.py
orbeckst: ok, thanks for explaining

src/alchemlyb/parsing/amber.py
```diff
@@ -65,8 +65,9 @@ def __init__(self, filename):
         self.filename = filename
         try:
             self.fileh = anyopen(self.filename, 'r')
-        except Exception as ex:  # pragma: no cover
+        except Exception as ex:
             logger.exception("Cannot open file %s", filename)
+            raise ex
```
orbeckst, line 70: IIRC, this can just be 
```suggestion
            raise
```
and it will re-raise the exception itself.

We can then simplify the line above, too.

src/alchemlyb/parsing/amber.py
```diff
@@ -65,8 +65,9 @@ def __init__(self, filename):
         self.filename = filename
         try:
             self.fileh = anyopen(self.filename, 'r')
-        except Exception as ex:  # pragma: no cover
+        except Exception as ex:
```
orbeckst, line 68: We don't need to alias as `ex`
```suggestion
        except:
```

If in doubt, less code is better ;-).

src/alchemlyb/parsing/amber.py
orbeckst: Let's raise an exception here. Use `ValueError` whenever something is wrong in the input file because we use `ValueError` in the NAMD parser in the same context.

Please add an appropriate test (copy & modify the test where you check for the logger messages — btw, nicely written tests! — and check `with pytest.raises(ValueError, match="The number of lambda windows read")`.

src/alchemlyb/parsing/amber.py
orbeckst: Let's raise an exception here. Use `ValueError` whenever something is wrong in the input file because we use `ValueError` in the NAMD parser in the same context. Please add a test.

orbeckst (changes requested):
Thank you for addressing my comments. 

I think we should raise exceptions in some of the cases where we currently warn and move on. Make these `logger.error` and raise an exception. There may be more cases in the parser where this is relevant. Thanks to your fine-grained test rewriting we can now test each of these cases properly and so increase the reliability of the AMBER parser.

Please also add entries to Changes in CHANGELOG for all the cases where we now raise an exception. Also add versionchanged — just in case there are users who will find that their workflow now raises exceptions.

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: Done and test added (thanks for the comment :-) )!

src/alchemlyb/parsing/amber.py
DrDomenicoMarson: Done and test added

orbeckst:
Let's get this PR merged. Add the CHANGES entries for the exceptions that you created here.

I made some direct edits, primarily use `logger.error()` instead of `logger.exception()` when not inside a try/except block. `logger.exception` will also log the currently active exception but in the case where _we_ raise an exception, there's no exception present yet.

orbeckst:
Each PR should be self contained so documenting what was done here is necessary in order to merge it.

Then please go ahead with a new PR related to #238 !

orbeckst:
Admittedly, I hadn't remembered #238 , otherwise I might not have suggested raising the exceptions here. But you already wrote the code and the tests so let's move forward with it. 

I am just waiting for CI to come back.

orbeckst (changes requested):
all looks great, just update CHANGES, please

DrDomenicoMarson:
Added the information to CHANGES!

orbeckst (approved):

PR #254, opened 2022-10-16
Clean up the iloc in the tests

### LINKED ISSUE #202: Switch tests to use `.loc` for all `delta_f_` comparisons
Consistent with how we instruct users in [our docs](https://alchemlyb.readthedocs.io/en/latest/estimators.html), we should [avoid using `iloc`](https://github.com/alchemistry/alchemlyb/pull/201#issuecomment-1170613385) and instead use `.loc` in our tests to specify which states to get free energy differences from.

xiki-tempula:
Fix #202

I have cleaned up most of the `iloc` in the test but there are a couple of tests that still use `iloc`. Most of them are parameterised tests where different datasets demand different column labels.

```python
    @pytest.mark.parametrize(('data', 'size'), [(gmx_benzene_dHdl(), 4001),
                                                (gmx_benzene_u_nk(), 4001)])
    def test_subsampling(self, data, size):
        assert len(self.slicer(data, series=data.iloc[:, 0])) <= size
```

In this case, the first column is always chosen which have different column name for different datasets.

Another case is 
```
    def get_delta_f(self, est):
        ee = 0.0

        for i in range(len(est.d_delta_f_) - 1):
            ee += est.d_delta_f_.values[i][i+1]**2
        return est.delta_f_.iloc[0, -1], ee**0.5
```
Where depending on the dataset, this could be `est.delta_f_[0.0][1.0]` or `est.delta_f_.loc[(0.0,0.0)][(1.0,1.0)]`


orbeckst:
I think the remaining iloc cases are fine. Could you add a short comment stating why these are ilocs, just so that what you found now is not forgotten?

xiki-tempula:
@orbeckst I have merged the master into this branch.

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -180,8 +180,9 @@ def _diag_sum(dataset):
             u_nk = extract_u_nk(filename, T=300)
 
             # Calculate the sum of diagonal elements:
-            for i in range(len(dataset['data'][leg])):
-                ds += u_nk.iloc[i][i]
+            for i, lambda_ in enumerate(u_nk.columns):
+                #18.6 is the time step
+                ds += u_nk.loc[i*186/10][lambda_].values[0]
```
orbeckst, line 185: If this is a proper `.loc`, can't we use
```suggestion
                ds += u_nk.loc[i*186/10, lambda_].values[0]
```

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -124,7 +124,7 @@ def test_u_nk_with_total_energy():
 
     # Check one specific value in the dataframe
     assert_almost_equal(
-        extract_u_nk(dataset['data']['AllStates'][0], T=300).iloc[0][0],
+        extract_u_nk(dataset['data']['AllStates'][0], T=300).loc[0][(0.0,0.0)].values[0],
```
orbeckst, line 127: PEP8 and using `.loc`
```suggestion
      extract_u_nk(dataset['data']['AllStates'][0], T=300).loc[0, (0.0, 0.0)].values[0],
```

src/alchemlyb/tests/parsing/test_gmx.py
```diff
@@ -161,7 +161,7 @@ def test_u_nk_without_energy():
...
```
orbeckst, line 164: see above

orbeckst (approved):
Thanks for the updates. 

I have a bunch of suggestions & PEP8 fixes. Primarily, can we avoid `df.loc[a][b]` and instead write `df.loc[a, b]` instead?

I am approving but please do at least the PEP8 formatting. Thank you!

src/alchemlyb/tests/parsing/test_gmx.py
xiki-tempula: This is not a proper `.loc`. 
the .loc[1.0] could in theory match both dataframe with row index (1.0, 0.0, 0.0) and (1.0).
But .loc[1.0, :] would only strictly match (1.0).

xiki-tempula:
@orbeckst For the PEP8, I guess I could just do a project-wide black, which is the same as the one for flamel?

orbeckst:
Can we do black in a separate PR? Reformatting hides any relevant changes. For this PR, do the few fixes manually and then we can blackify in a separate PR.

xiki-tempula:
@orbeckst Would it be easier if I merge this PR as it is. Then do a black PR? 
For the .loc[0][0] related things, I cannot change them.
For the space after the `,`, black will automatically sort them out.

src/alchemlyb/tests/parsing/test_gmx.py
orbeckst: Ok, after looking at the data I see that this is a multiindex. But in this case wouldn't be
```
ds += u_nk.loc[i*186/10].loc[lambda_].values[0]
```
be cleaner?

(I have no idea how to get something like ".loc[lambda_, (0.0, 0.0)]" instead of the values[0] but I find pandas indexing confusing...)

orbeckst (approved):
that's fine then; run black later

pandas indexing is a bit confusing...

src/alchemlyb/tests/parsing/test_gmx.py
xiki-tempula: Yes, but we cannot get lambda_ beforehand and the lambda_ could be (0.0, 0.0) or (0.0, 0.0, 0.0), which is why for some of them, I could only do `iloc`.
In some sense, we could get them by analysing the dataframe but that would make the unit test too complicated so I think the current form is a good balance.

PR #255, opened 2022-10-19
update to deploy workflow

orbeckst:
- use Python 3.9 instead of 3.8
- update to release/v1 of pypa action (instead of deprecated @master)


xiki-tempula (approved):
lgtm

PR #260, opened 2022-10-29
Default the plotting unit to None

### LINKED ISSUE #247: Default the unit of the plotting function to None

xiki-tempula:
Fix #247
Default the unit in `plot_dF_state` , `plot_convergence` and `plot_ti_dhdl` to None.
Remove the array input in `plot_convergence`.

I have fixed the issue here @DrDomenicoMarson

### DIFF
--- a/src/alchemlyb/visualisation/convergence.py
+++ b/src/alchemlyb/visualisation/convergence.py
@@ -5,37 +5,35 @@ import numpy as np
 
 from ..postprocessors.units import get_unit_converter
 
-def plot_convergence(*data, units='kT', final_error=None, ax=None):
+def plot_convergence(dataframe, units=None, final_error=None, ax=None):
     """Plot the forward and backward convergence.
 
     The input could be the result from
-    :func:`~alchemlyb.convergence.forward_backward_convergence` or it could
-    be given explicitly as `forward`, `forward_error`, `backward`,
-    `backward_error`.
+    :func:`~alchemlyb.convergence.forward_backward_convergence` or
+    :func:`~alchemlyb.convergence.fwdrev_cumavg_Rc`. The input should be a
+    :class:`pandas.DataFrame` which has column `Forward`, `Backward` and
+    :attr:`pandas.DataFrame.attrs` should compile with :ref:`note-on-units`.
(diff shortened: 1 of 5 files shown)


src/alchemlyb/visualisation/convergence.py
```diff
@@ -5,37 +5,35 @@
...
 
     The input could be the result from
-    :func:`~alchemlyb.convergence.forward_backward_convergence` or it could
-    be given explicitly as `forward`, `forward_error`, `backward`,
-    `backward_error`.
-
-    `forward`: A list of free energy estimate from the first X% of data,
-    where `forward_error` is the corresponding error.
+    :func:`~alchemlyb.convergence.forward_backward_convergence` or
+    :func:`~alchemlyb.convergence.fwdrev_cumavg_Rc`. The input should be a
```
orbeckst, line 13: (from other PR on Rc but that's fine)

orbeckst (changes requested):
Looking good but please add entry in CHANGES under Changes

xiki-tempula:
@orbeckst Thanks for the review.
I have added the change log

> Is there a test that shows that (1) None works and that (2) changing units works?

I have added this as well.

orbeckst (approved):

orbeckst:
thanks!

Issue #208, opened 2022-07-23
release 1.0


orbeckst:
Create a 1.0 release (we have a reasonably stable API). 

* This would be the last release that supports pymbar 3 (see #205 ).
* Should contain the ABFE workflow.

[Release management](https://github.com/alchemistry/alchemlyb/wiki/Developer-Guide#release-management)
- [x] updated CHANGES
- [x] tagged 1.0.0
- [x] checked testpypi version
- [x] created release on GitHub
- [x] created conda package (reviewed and merged the cf-regro commit to the conda-forge repo for the conda package)
- [x] announced in forum

orbeckst:
@xiki-tempula @DrDomenicoMarson what should be accomplished before we can release a 1.0? Those issues/PRs should be added to the [1.0.0 milestone](https://github.com/alchemistry/alchemlyb/milestone/7).

orbeckst:
Sounds sensible. Could you please adjust the milestone tags as necessary, @xiki-tempula ?

DrDomenicoMarson:
I think we need to have a somewhat definitive solution/workaround for issue #242 before 1.0.0, I'm waiting for a definitive response from AMBER developers. Tomorrow I'll better explain the situation in the relevant thread, and I'll try to propose a reliable solution!

xiki-tempula:
@DrDomenicoMarson The 1.0.0 is just to make the API stable. For #242, if it is too difficult, we could just do a 1.0.1. But it would be really helpful if this could be done before 1.0.0.

xiki-tempula:
I'm fine with a 1.0 release. As usual, I would imagine that 1.0.1 release will soon follow.

orbeckst:
The [1.0.0 release](https://github.com/alchemistry/alchemlyb/releases/tag/1.0.0) is now available. It should be on conda-forge within a day or so.

@DrDomenicoMarson do you have a twitter handle so that I can give you some credit in the announcement tweet?

DrDomenicoMarson:
Wow, that was quick :-)
My Twitter handle is DomenicoM87!

orbeckst:
Thanks. I am still waiting for conda-forge https://github.com/conda-forge/alchemlyb-feedstock/ (there's an automated process that creates a PR from the PyPi package but it typically has a couple of hours backlog)

orbeckst:
Manually working on the cf package in https://github.com/conda-forge/alchemlyb-feedstock/pull/6

[...]

PR #261, opened 2022-10-31
finalize 1.0.0

orbeckst:
- for #208 
- CHANGES
- docs
- additional tests


src/alchemlyb/convergence/convergence.py
```diff
@@ -216,16 +223,25 @@ def fwdrev_cumavg_Rc(series, precision=0.01, tol=2):
...
-    modified based on Shujie Fan's (@VOD555) work.
-   
-   Please cite [Fan2021]_ when using this function.
-   
+    Notes
+    -----
+    This function computes :math:`R_c` from `equation 16`_ from [Fan2021]_.
+    The code is modified based on Shujie Fan's (@VOD555) work. Zhiyi Wu
+    (@xiki-tempula) improved the performance of the original algorithm by
+    pre-computing block averages for fractional increments of `precision`.
```
xiki-tempula, line 231: Shujie Fan's (@VOD555) work has already implemented the pre-computing block averages. His work uses the normal python for loop and I have improved the performance by vectorise it using numpy array.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -169,22 +191,27 @@ def u_nk2series(df, method='dE'):
...
 
 
 @pass_attrs
-def dhdl2series(df, method=''):
+def dhdl2series(df, method='all'):
     """Convert a dhdl DataFrame to a series for subsampling.
 
+    The series is generated by summing over all energy components (axis 1 of
+    `df`), as for ``method='all'`` in :func:`u_nk2series`. Normally, `df` only
+    contains a single energy component.
```
xiki-tempula, line 206: I think it might not be very good to have `Normally` in

> Normally, `df` only contains a single energy component.

If one use split protocol in Gromacs, the dhdl will always have more than one column and using the phrase `Normally` might indicate that having more than one column in dHdl is abnormal. I think it might be good to change it to 

> `Most `df` only contains a single energy component.

xiki-tempula (changes requested):
Many thanks for finalising the release just some minor doc comments.

orbeckst:
Thank you for the clarifications, I will change the text. I’m very glad that you looked over everything carefully.

src/alchemlyb/convergence/convergence.py
orbeckst: thanks, reworded

src/alchemlyb/preprocessing/subsampling.py
orbeckst: thanks, reworded

PR #266, opened 2022-11-01
small fixes for 1.0.1

orbeckst:
Invariably, we missed something for the 1.0.0 release #265 ... so here we can collect little things like doc fixes that don't need their own issue.

Anyone should feel free to push to this branch or directly edit files on the branch.


setup.py
```diff
@@ -31,6 +31,7 @@
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
```
orbeckst, line 34:
```suggestion
```
Leave py 3.11 to PR #269

xiki-tempula (approved):
LGTM Thanks.

PR #275, opened 2022-11-20
Slicing will no longer drop the NaN rows

### LINKED ISSUE #274: NAMD - decorrelation and equilibration detection fails in v0.7 and 1.0
Recent updates (to preprocessing.py, I believe) have broken our scripts' ability to use decorrelation and equilibration detection for NAMD output. I have attached an MWE that includes some sample data from alchemtest. The most recent version that works for these test cases is Alchemlyb 0.6.0. I have included sample logs for runs using 0.6, 0.7, and 1.0

Finally, it looks like decorrelation should work on the entire dataframe based on the documentation, but we have had to separate the dataframe by fep-lambda. Otherwise, decorrelate_u_nk returns just fep-lambda=0.

Any clarification or advice would be helpful!

MWE: [alchemlyb_crash_MWE.zip](https://github.com/alchemistry/alchemlyb/files/10024970/alchemlyb_crash_MWE.zip)

xiki-tempula:
Fix #274 
The Pre-processing function slicing will not drop NaN rows.


xiki-tempula:
@orbeckst I don't know why RTD failed but it works on my local computer

orbeckst (approved):
Not dropping NaN makes sense. We didn't even state that we were doing it so this is a fix to remove undocumented and potentially erroneous behavior.

jhenin:
Thank you @xiki-tempula and @orbeckst !

PR #277, opened 2022-12-04
fix rtd

xiki-tempula:
https://github.com/readthedocs/readthedocs.org/issues/9776


orbeckst (approved):
Thank you for investigating and solving the issue.

PR #280, opened 2022-12-05
Blackfy the code

xiki-tempula:
Refactor the code according to black and add black check to the ci.


orbeckst (approved):
ok

(I am not 100% on board with _all_ of blacks formatting but I don't want to worry about it... so all fine)

xiki-tempula:
@orbeckst If one doesn't want to blackfy certain regions of code, one could use something similar to 'param: no cover' to allow black to ignore it.

orbeckst:
Yeah, it's ok. I think I am ok with the trade-off between not having to think about formatting ever again and making minor decisions about preferred/marginally more legible indentation etc. It's fine.

PR #281, opened 2022-12-07
refactor ABFE tests

xiki-tempula:
Cluster the tests based on their function.

PR #282, opened 2022-12-07
Fix #276

### LINKED ISSUE #276: ABFE workflow cannot handle 0 or False for forwrev
if I don't need to do convergence analysis, I will usually do `ABFE(forwrev=None)` and if I want to do convergence analysis, I will do `ABFE(forwrev=10)`, these two ways of using ABFE are all fine. However, if someone does `ABFE(forwrev=0)` or `ABFE(forwrev=False)`, the workflow will crash.

xiki-tempula:
Merge after #281
Fix #276


orbeckst:
please resolve conflicts

orbeckst (approved):
Looking good — simple fix. I assume that there's a test somewhere but that was difficult to see because it would need rebasing.

Resolve the conflicts and then squash-merge.

PR #289, opened 2022-12-10
Raise Exception when no file has been matched in ABFE workflow

xiki-tempula:
Related to https://github.com/alchemistry/flamel/issues/24. I noticed that when no file has been matched by the directory, prefix and suffix, no useful error message is being displayed. This PR should fix this issue.


orbeckst (approved):
sensible fix, lgtm

PR #290, opened 2022-12-11
Make output directory when it didn't exist for workflow

xiki-tempula:
In https://github.com/alchemistry/flamel/issues/24, it is reported that when the output directory didn't exist, the flow will crash. This PR make it that if the output directory didn't exist, it will be created.


orbeckst (approved):
lgtm

PR #300, opened 2023-03-18
Fix the dE method for AMBER

### LINKED ISSUE #299: The dE method in decorrelating the u_nk data is not actually dE
Fro Gromacs each column in the u_nk is the dE between the reduced potential of the current lambda (the lambda of the column) and the reduced potential of the lambda that is being simulated. So for the dE method in decorrelation, we just use the column after or before the current lambda as it is already dE.

However, this assumption is not always the case particularly when one wants to use the function to decorrelate the openmmtools output where one could just put in a dataframe where each column is the reduced potential of that lambda window. In this case, the decorrelation will be based on the reduced potential instead of dE.

xiki-tempula:
Fix #299

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,7 +17,7 @@ The rules for this file:
 
   * unreleased
 
-Changes
+Fixes
   - Change the dE method in u_nk2series to use the difference between two
     lambda columns (issue #299, PR #300).
 
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,15 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" https://semver.org
 
 ------------------------------------------------------------------------------
+*/*/* xiki-tempula,
+
+  * unreleased
+
+Changes
+  - Change the dE method in u_nk2series to use the difference between two
+    lambdas (issue #299, PR #268).
+
+
 12/12/2022 xiki-tempula, orbeckst
 
   * 2.0.0
--- a/CHANGES
+++ b/CHANGES
@@ -19,7 +19,7 @@ The rules for this file:
 
 Changes
   - Change the dE method in u_nk2series to use the difference between two
-    lambdas (issue #299, PR #268).
+    lambda columns (issue #299, PR #300).
 
 
 12/12/2022 xiki-tempula, orbeckst
--- a/CHANGES
+++ b/CHANGES
@@ -18,8 +18,9 @@ The rules for this file:
   * unreleased
 
 Fixes
-  - Change the dE method in u_nk2series to use the difference between two
-    lambda columns (issue #299, PR #300).
+  - Fix the dE method in u_nk2series to use the difference between two
+    lambda columns instead of using the next lambda column or the previous
+    column for the last window (issue #299, PR #300).
 
 
 12/12/2022 xiki-tempula, orbeckst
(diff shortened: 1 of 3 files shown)


CHANGES
```diff
@@ -13,6 +13,15 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" https://semver.org
 
 ------------------------------------------------------------------------------
+*/*/* xiki-tempula,
+
+  * unreleased
+
+Changes
```
orbeckst, line 20: Is this really a change or shouldn't this be listed as a Fix because it now does what is advertised?

orbeckst (changes requested):
I need some clarification: From the comments in the issue and PR, it sounds as if GROMACS data were correctly processed while others were not — correct me if I misunderstood. What does happen now to GROMACS data after this change?

Also, if it's a Change then there should be a versionchanged in u_nk2series(). However, it sounds more like a fix to me to bring the code in line with the documentation. What do you think it should be classified as?

xiki-tempula:
> I need some clarification: From the comments in the issue and PR, it sounds as if GROMACS data were correctly processed while others were not — correct me if I misunderstood. What does happen now to GROMACS data after this change?

GROMACS data should not be changed. The numerical tests with regard to gromacs is not changed.

For a simulation with lambda=0.5, the Gromacs has
| Lambda | 0  | 0.5 | 1.0 |
|--------|----|-----|-----|
| 1      | -1 | 0   | 1   |
Previously, we just take `data[lambda=1.0] = 1`
We now do `data[lambda=1.0] - data[lambda=0.5] = 1 - 0 = 1`

CHANGES
xiki-tempula: versionchanged added

xiki-tempula:
Sorry, I just noticed that I'm wrong.
For Gromacs, it is true that the column corresponding to the current lambda is 0. However, when we read in the reduced potential, it is the dE corrected by the pV, so this changes the gromacs result as well but to the good side.

CHANGES
```diff
@@ -13,6 +13,15 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" https://semver.org
 
 ------------------------------------------------------------------------------
+*/*/* xiki-tempula,
+
+  * unreleased
+
+Fixes
+  - Change the dE method in u_nk2series to use the difference between two
+    lambda columns (issue #299, PR #300).
```
orbeckst, line 22: Can you be more specific what *bug* was fixed — instead of the word "Change", something like "fixed <description of problem> to now use .... (note: will lead to different results than before) (issue ....)"

If I am a user and look through CHANGES I would want to understand if anything affects me.

src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -153,6 +153,9 @@ def u_nk2series(df, method="dE"):
 
 
     .. versionadded:: 1.0.0
+    .. versionchanged:: 2.0.1
+       The `dE` method computes the difference between the current lambda
+       and the next lambda (previous lambda for the last window).
```
orbeckst, line 158: add "... instead of ..." so that people have an idea what behavior changed

orbeckst (changes requested):
Thank for the explanation and digging into what effect the fix will have.

Could you please improve the documentation a little bit? I just want to give users a fair chance to understand where results are changing.

orbeckst (approved):

orbeckst:
Thanks @xiki-tempula !

PR #304, opened 2023-04-03
TI with gaussian quadrature

hl2500:
Added an optional gaussian quadrature functionality to estimate the TI integral. 

This functionality could be triggered when initializing: `TI(gauss_qua=True)`

The lambda values and weights were referred the Amber manual (page 493): http://ambermd.org/doc12/Amber22.pdf

The formula used to estimate the free energy and standard errors were referred the the Amber manual (eq. 25.2) and this literature (eq. 3): https://pubs.acs.org/doi/10.1021/acs.jcim.2c01052?ref=pdf, respectively.

The final two dataframes with free energy values and errors are upper triangluar matrix, because the values are cumulative sum upto each lambda state rather than differences. The diagonal values are the estimation at each lambda.

I have tested 12 lambdas and 9 lambdas with Amber one-step method to estimate the free energy and error of each leg in the thermodynamic cycle. The results are comparable to the ones estimated with trapezoidal rule. The other number of lambdas (1, 2, 3, 5, 7) should work as well.


src/alchemlyb/estimators/ti_.py
```diff
@@ -30,14 +30,22 @@ class TI(BaseEstimator, _EstimatorMixOut):
     dhdl : DataFrame
         The estimated dhdl of each state.
 
+    _gq_delta_f_ : DataFrame, optional
+        The estimated dimensionless cumulative sum of free energy up to each state 
+        with gaussian quadrature.
+
+    _gq_d_delta_f_ : DataFrame, optional
+        The estimated statistical uncertainty (one standard deviation) in 
+        the cumulative sum of free energy up to each state with gaussian quadrature
```
orbeckst, line 39: Will users want to access these data structures? If yes, no underscore and document them as you do.

If they are private attributes then they don't need to be documented.

src/alchemlyb/estimators/ti_.py
```diff
@@ -30,14 +30,22 @@ class TI(BaseEstimator, _EstimatorMixOut):
...
 
     .. versionchanged:: 1.0.0
        `delta_f_`, `d_delta_f_`, `states_` are view of the original object.
```
orbeckst, line 43: You'd need to add a versionchanged.

src/alchemlyb/estimators/ti_.py
```diff
@@ -30,14 +30,22 @@ class TI(BaseEstimator, _EstimatorMixOut):
...
 
     """
 
-    def __init__(self, verbose=False):
+    def __init__(self, verbose=False, gauss_qua=False):
```
orbeckst, line 46: I prefer written out keywords, eg. `quadrature`.

src/alchemlyb/estimators/ti_.py
```diff
@@ -30,14 +30,22 @@ class TI(BaseEstimator, _EstimatorMixOut):
...
         self.verbose = verbose
+        self.gauss_qua = gauss_qua
```
orbeckst, line 48: `use_gaussian_quadrature` to be clear

src/alchemlyb/estimators/ti_.py
```diff
@@ -166,3 +177,129 @@ def separate_dhdl(self):
...
+        ----------
+        dHdl : DataFrame
+        means : DataFrame
+        variances: DataFrame
+        """
+        # extract the lambda values used in the simulations
+        lambdas = means.reset_index()[means.index.names[:]].iloc[:].values.sum(axis=1)
+        num_lambdas = len(lambdas)
+        # check if the lambda values used match the common ones suggested in Amber manual and assign cooresponding weights
+        if num_lambdas==1:
```
orbeckst, line 196: Avoid code repetition and that many elif statements. Make the code much more compact and extendable by storing the special points and weights in a data structure
```
special_points = {1:   [[0.5], [1.0]], 
                             2:  [[0.21132, 0.78867], [0.5, 0.5]],
                             3:  [[0.1127, 0.5, 0.88729],  [0.27777, 0.44444, 0.27777]]
                             ...}
``` 
and have one code path that pulls its points from the `special_points`.

src/alchemlyb/estimators/ti_.py
```diff
@@ -166,3 +177,129 @@ def separate_dhdl(self):
...
+            try:
+                reference_lambdas = np.round([0.5], 4)
+                assert (lambdas == reference_lambdas).all()
```
orbeckst, line 199: assert is not the correct way to check input. If there's anything wrong with user data you need to raise an exception, probably a ValueError.

Furthermore, *never* use `==` to check floating point number equality. Instead use `np.allclose()` or similar.

orbeckst (changes requested):
I had a really quick look. In addition to my comments you also need to
- [ ] add tests that cover all your code
- [ ] add documentation (example to the TI page, more details to the TI class) and cite any relevant papers
- [ ] add an entry to CHANGES
- [ ] add yourself to AUTHORS

Addressing the changes is no guarantee that the PR is accepted; addressing them is a necessary but not sufficient condition. I'd also like to have @xiki-tempula look at it and @mrshirts if he can spare some time (mainly on the "is it worth it", based on the papers you cited)

xiki-tempula (changes requested):
I think it would be good to add a new estimator in parallel to the current TI instead of adding it  to TI. If you need something that is in TI, might be good to abstract out a BaseTI class.

orbeckst:
Sorry, I don't have much time this/next week to review. I hope @xiki-tempula can have a look. I agree with the suggestion to make it a separate estimator instead of changing standard TI.

hl2500:
Thank you for the comments @orbeckst @xiki-tempula I may need some help to add tests to cover all code. I didn't find a testset in alchemtest for Amber gaussian quadrature (with suggested lambda values). Do I also need to upload an example to alchemtest for the purpose of testing and adding an example on TI documentation page? Thanks!

CHANGES
```diff
@@ -13,6 +13,15 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" https://semver.org
 
 ------------------------------------------------------------------------------
+06/04/2023 hl2500
```
xiki-tempula, line 16: `*/*/* hl2500`

As we don't know when to release the new version yet

CHANGES
```diff
@@ -13,6 +13,15 @@ The rules for this file:
...
+
+  * 2.0.1
```
xiki-tempula, line 18: `* Unreleased`

src/alchemlyb/estimators/ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,165 @@
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator
+
+from .base import _EstimatorMixOut
+
+
+class TI_gq(BaseEstimator, _EstimatorMixOut):
+    """Thermodynamic integration (TI) with gaussian quadrature estimation.
+
```
xiki-tempula, line 10: Add a bit of background of how is this different from the TI

src/alchemlyb/estimators/ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,165 @@
...
+        dimensionless free energy differences.
+
+    states_ : list
+        Lambda states for which free energy estimation were obtained.
+
+    dhdl : DataFrame
+        The estimated dhdl of each state.
+
+
+    .. versionchanged:: 2.0.0
```
xiki-tempula, line 34: You don't need this. Do `versionadded` instead

src/alchemlyb/estimators/ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # sort by state so that rows from same state are in contiguous blocks,
+        # and adjacent states are next to each other
+        dHdl = dHdl.sort_index(level=dHdl.index.names[1:])
+
+        # obtain the mean and variance of the mean for each state
+        # variance calculation assumes no correlation between points
+        # used to calculate mean
+        means = dHdl.groupby(level=dHdl.index.names[1:]).mean()
+        variances = np.square(dHdl.groupby(level=dHdl.index.names[1:]).sem())
```
xiki-tempula, line 63: For the part that is the same in TI and TI_GQ, abstract a TIBase class and put them there. We don't want code duplication.

src/alchemlyb/estimators/ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # extract the lambda values used in the simulations
+        lambdas = means.reset_index()[means.index.names[:]].iloc[:].values.sum(axis=1)
```
xiki-tempula, line 66: Is this safe for the dataframe that has more than one lambda values?

src/alchemlyb/estimators/ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,165 @@
...
+        num_lambdas = len(lambdas)
+        # suggested lambda values corresponding weights 
+        special_points = {1: [[0.5], [1.0]], 2: [[0.21132, 0.78867], [0.5, 0.5]],
```
xiki-tempula, line 69: It is better to use dispatch than a bunch of if else

src/alchemlyb/estimators/ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,165 @@
...
+            index=variances.index.values,
+        )
+        self._states_ = means.index.values.tolist()
+        self._delta_f_.attrs = dHdl.attrs
+        self._d_delta_f_.attrs = dHdl.attrs
+        self.dhdl.attrs = dHdl.attrs
+
+        return self
+
+    def separate_dhdl(self):
```
xiki-tempula, line 123: Is this the same as the one in TI?

xiki-tempula (changes requested):
Thanks for this PR. It looks reasonable, I have had a quick first pass.

- We need some tests to test this method, you need to add them to alchemtest.
- Add documentation to explain what is TI_GQ and when and why do we want it.
- Abstract out the code that is the same between TI and TI_GQ as TIBase and inherit from it.

hl2500:
Thank you so much for the suggestions. I have add two testsets to alchemtest (PR #87) and made some changes. @orbeckst @xiki-tempula

src/alchemlyb/estimators/ti_gaussian_quadrature.py
hl2500: I modified this part slightly to use the separate_dhdl(self) function.

src/alchemlyb/estimators/ti_gaussian_quadrature.py
hl2500: Now this function is different to support multi-lambda situation.

src/alchemlyb/estimators/ti_gaussian_quadrature.py
hl2500: I changed this part to support multi-lambda situation. However, scaling multiple lambdas at the same time is not supported

xiki-tempula:
@hl2500 Thanks for the new test files updated in alchemtest. Do you mind test them in this PR? So I could know what are the test for and how they have been used.

hl2500:
@xiki-tempula Okay. Could I ask how to implement test in this PR? Thanks.

xiki-tempula:
I think if you just change this line https://github.com/alchemistry/alchemlyb/blob/master/.github/workflows/ci.yaml#L58
to
python -m pip install git+https://github.com/hl2500/alchemtest.git@Add_gaussian_quadrature_test
You should be able to get this in the CI.

hl2500:
@xiki-tempula I have changed this line and some of the ubuntu tests failed. Could I ask what should I do? Thanks.

xiki-tempula:
Just Codecov acting weird. It is fine now.

hl2500:
@xiki-tempula Thanks, could I ask what other tests I need to do?

xiki-tempula:
@hl2500 Thanks for the contribution, so the test should be in a format that is similar to https://github.com/alchemistry/alchemlyb/blob/master/src/alchemlyb/tests/test_ti_estimators.py, which is for the TI estimator. Which shall test if the result is numerically good.
The other thing that you need to test is if the test is covering all of the code, you could check the coverage here https://github.com/alchemistry/alchemlyb/pull/304/checks?check_run_id=12878246768, we would like to pump it as high as possible.

hl2500:
@xiki-tempula Thank you for the information! Some tests have been added to improve the coverage.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+        dimensionless cumulative free energies.
+
+    states_ : list
+        Lambda states for which free energy estimation were obtained.
+
+    dhdl : DataFrame
+        The estimated dhdl of each state.
+
+
+    .. versionadded:: 2.0.0
```
xiki-tempula, line 37: 2.1.0

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+        """
+
+        # sort by state so that rows from same state are in contiguous blocks,
+        # and adjacent states are next to each other
+        dHdl = dHdl.sort_index(level=dHdl.index.names[1:])
+
+        # obtain the mean and variance of the mean for each state
+        # variance calculation assumes no correlation between points
+        # used to calculate mean
+        self.means = dHdl.groupby(level=dHdl.index.names[1:]).mean()
```
xiki-tempula, line 65: Why do you need `self.means` and `self.variances`? I mean instead of just means and variances? Or even as private variable.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+        
+        self.dhdl = self.means
+        self.dhdl.attrs = dHdl.attrs
+        self._states_ = means_new.index.values.tolist()
+        self._delta_f_.attrs = dHdl.attrs
+        self._d_delta_f_.attrs = dHdl.attrs
+
+        return self
+
+    def separate_dhdl(self):
```
xiki-tempula, line 122: Do you need this as a public function? We have `TI.separate_dhdl()` so we would like this to have a different name.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+        self.variances = np.square(dHdl.groupby(level=dHdl.index.names[1:]).sem())
+        # suggested lambda values and corresponding weights 
+        special_points = self.get_quadrature_points()
+        weights = []
+        # check if the lambdas in the simulations match the suggested values
+        lambda_list, means_list, variances_list = self.separate_dhdl()
+        for lambdas in lambda_list:
+            num_lambdas = len(lambdas)
+            if (num_lambdas not in special_points) or (not np.allclose(lambdas, special_points[num_lambdas][0], rtol=0.1)):
+                raise ValueError('The lambda values for gaussian quadrature are not supported, please use trapezoidal rule instead.')
```
xiki-tempula, line 75: Might mention in the doc that this method only supports some specific given lambda schedules.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+            weights.extend(special_points[num_lambdas][1])
```
xiki-tempula, line 76: This `[1]` is a bit confusing. Might just make `special_points` as a class attribute in the format of
`special_points = {1: {'lambdas': [], 'weights': []}, 2: {'lambdas': [], 'weights': []},}`

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+        )
+        
+        # yield standard deviation d_delta_f_ between each state
+        self._d_delta_f_ = pd.DataFrame(
+            np.sqrt(ad_delta),
+            columns=variances_new.index.values,
+            index=variances_new.index.values,
+        )
+        
+        self.dhdl = self.means
```
xiki-tempula, line 114: I think you don't really need `self.means`? just call it `self.dhdl`?

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
xiki-tempula: Might make this into a static function that takes `self.means` as input and return the results.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+        if len(self.means.index.names) == 1:
+            name = self.means.columns[0]
+            lambda_list.append(self.means.index)
+            dhdl_list.append(self.means[name])
+            variance_list.append(self.variances[name])
+
+        else:
+            # simultanouesly scaling of multiple lambda types are not supported
+            if (((0.0 < lambdas) & (lambdas < 1.0)).sum(axis=1) > 1.0).any():
+                raise ValueError('The lambda values for gaussian quadrature are not supported, please use trapezoidal rule instead.')
```
xiki-tempula, line 163: Please make this error message more clear. 
Also `(((0.0 < lambdas) & (lambdas < 1.0)).sum(axis=1) > 1.0).any()` is a bit confusing to read. Could you give some failed test examples in the test so I could see when this will trigger?

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,228 @@
...
+            for i in range(len(l_types)):
+                # obtain the lambda points between 0.0 and 1.0
```
xiki-tempula, line 165: How about the value that is outside the range? Why would you expect to get lambda that is outside the range?

src/alchemlyb/tests/test_ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    def X_delta_f(self, request):
+        get_dHdl, E, dE = request.param
+        return request.getfixturevalue(get_dHdl), E, dE
+
+
+def test_TI_GQ_separate_dhdl_multiple_column(ethanol):
+    dHdl = ethanol
+    estimator = TI_GQ().fit(dHdl)
+    assert isinstance(estimator.separate_dhdl()[0], list)
```
xiki-tempula, line 80: It is good to test this estimator.separate_dhdl() but might be more clear to make this into a class and test various aspects. In theory one test should only be testing one thing.

src/alchemlyb/tests/test_ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,168 @@
...
+    dHdl.set_index("bound-lambda", append=True, inplace=True)
+    estimator = TI_GQ().fit(dHdl)
+    assert all([isinstance(dhdl, pd.Series) for dhdl in estimator.separate_dhdl()[1]])
+    assert [len(dhdl) for dhdl in estimator.separate_dhdl()[1]] == [12]
+
+
+def test_TI_GQ_not_quadrature_points(benzene_VDW):
+    """The test for the case where the simulation lambdas are not quadrature points"""
+    dHdl = benzene_VDW
+    try:
```
xiki-tempula, line 115: Use `pytest.raise` to catch the error

src/alchemlyb/tests/test_ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,168 @@
...
+
+def test_TI_GQ_multi_lambda_scaling(ethanol_Coulomb):
+    """The test for the case where multiple lambdas are scaled simultaneously (not supported)"""
+    dHdl = ethanol_Coulomb
+    # change the second lambda from all zeros to the same as the first lambda
+    dHdl.reset_index(inplace=True)
+    dHdl["vdw-lambda"] = dHdl["coul-lambda"]
+    dHdl.set_index(["time", "coul-lambda", "vdw-lambda"], inplace=True)
+    try:
+        TI_GQ().fit(dHdl)
```
xiki-tempula, line 129: Use `pytest.raise` to catch the error

xiki-tempula (changes requested):
Thanks very much for this PR. It is taking shape. Sorry that I'm not too familiar with the estimator so some part of the code is not too clear to me. Do you mind address some of my questions, please? Thank you.

It might also be good to 

- Add a section in the doc that explains what this new estimator does and when can or cannot one use it.
- Add a test to check if one is getting similar result from TI and TI_QC.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
hl2500: I have changed this to a static function with a different name and it takes means and variances directly.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
hl2500: I have changed the error message.
`(((0.0 < lambdas) & (lambdas < 1.0)).sum(axis=1) > 1.0).any()` is to make sure that in the case of multiple lambda, only one lambda could be scaled at a time, for example (0.0469, 0.0), (0.2308, 0.0), ..., (0.9531, 0. 0). Otherwise, the sum would be greater than 1 and trigger that error.
There is a test example (`test_TI_GQ_multi_lambda_scaling(ethanol_Coulomb)`) where a case was created artificially by scaling both lambdas at the same time, for example (0.0469, 0.0469), (0.2308, 0.2308), ..., (0.9531, 0.9531). This lambda schedule might be problematic when applied in real simulations. The error would be caught by `pytest.raise`.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
hl2500: Since gaussian quadrature lambdas are between 0 and 1, only lambda values greater than 0 and less than 1 would be extracted and compared with the ones in `special_points` dictionary. The lambda values 0 and 1 would be ignored, for example, even if the simulation is performed under the lambdas `(0.0, 0.0469, 0.2308, 0.5, 0.7692, 0.9531, 1.0)`, only `(0.0469, 0.2308, 0.5, 0.7692, 0.9531)` would be compared with the suggested values (which matches the `special_points[5]`) and be used to estimate the free energy. 
On the other hand, if the simulations are performed at `(0.0, 0.2, 0.4, 0.6, 0.8, 1.0)`, only `0.2, 0.4, 0.6, 0.8)` would be extracted and compared against the `sepcial_points` and there would be an error, because they are not gaussian quadrature points.

hl2500:
@xiki-tempula Thank you so much for the suggestions! Please feel free to let me know if any other changes are needed.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,220 @@
...
+
+        if len(means.index.names) == 1:
+            name = means.columns[0]
+            lambda_list.append(means.index)
+            dhdl_list.append(means[name])
+            variance_list.append(variances[name])
+
+        else:
+            # simultanouesly scaling of multiple lambda types are not supported
+            if (((0.0 < lambdas) & (lambdas < 1.0)).sum(axis=1) > 1.0).any():
```
xiki-tempula, line 166: I'm a bit confused as to when this would be triggered. So I would imagine that for lambda `coul` and `vdw`.
You would want to accept

|coul|vdw|
|-|---|
|0|0.1|
|0|0.2|
|0|...|
|0|0.9|

Or

|coul|vdw|
|-|---|
|1|0.1|
|1|0.2|
|1|...|
|1|0.9|

Or

|coul|vdw|
|-|---|
|0.1|0.1|
|0.2|0.2|
|...|...|
|0.9|0.9|

And you want to reject

|coul|vdw|
|-|---|
|0.2|0.1|
|0.5|0.2|
|...|...|
|1|0.9|
?

xiki-tempula (changes requested):
Thanks for addressing the comments. We are getting there just a couple of cosmetic comments.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
hl2500: @xiki-tempula Thanks for the comment. Could I double check with you if the coul and vdw columns are independent from each other? For example, if we have a table of the means like:

| coul-lambda | vdw-lambda | coul-dHdl-means | vdw-dHdl-means |
| ------------  | ------------  | -----------  | -----------  | 
| 0.0469 | 0.0469 | 25.71 | -12.65 |   
| 0.2308 |  0.2308 | 17.67 | -3.11 |
| 0.5 | 0.5 | 8.77 | 3.15 |
| 0.7692 | 0.7692 | 3.17 | 5.27 |
| 0.9531 | 0.9531 | 0.56 | 5.36 |

Could we do the estimation separately? Like:
`∆G_coul = W_0.0469 * 25.71 + W_0.2308 * 17.67  + W_0.5 * 8.77 + W_0.7692 * 3.17 + W_0.9531 * 0.56`
`∆G_vdw = W_0.0469 * (-12.65) + W_0.2308 * (-3.11)  + W_0.5 * 3.15 + W_0.7692 * 5.27 + W_0.9531 * 5.36 `
`∆G = ∆G_coul + ∆G_vdw`

If so, perhaps we don't need this check.

xiki-tempula: > Could I double check with you if the coul and vdw columns are independent from each other? 

For Gromacs, this is the case. If you check the TI code then it just add them together. I think this check might not be necessary.

hl2500: Okay, sounds good. I will remove it then.

xiki-tempula:
@hl2500 Sorry for this late comment. I was playing with this estimator on some datasets and found a problem.
The current layout for TI, BAR, MBAR is that for delta_f_ of three lambda [0, 0.5, 1.0].
The [0,0] is zero, and the [0,1] is the difference between the zeroth element in the lambda list [0] and the first element in the lambda list [0.5], the [0,2] is the difference between the zeroth element in the lambda list [0] and the first element in the lambda list [1.0].

Do you mind making this estimator follow this convention as well? Thank you.

src/alchemlyb/tests/test_ti_gaussian_quadrature.py
```diff
@@ -0,0 +1,227 @@
...
+
+@pytest.fixture
+def ethanol_VDW(gmx_ethanol_VDW_dHdl):
+    dHdl = alchemlyb.concat(gmx_ethanol_VDW_dHdl)
+    return dHdl
+
+
+@pytest.fixture
+def ethanol(ethanol_Coulomb, ethanol_VDW):
+    dHdl = alchemlyb.concat([ethanol_Coulomb, ethanol_VDW])
```
xiki-tempula, line 27: Sorry, it could be that I'm doing something wrong, I'm testing this on a set that has two lambda values, such as the ethanol that you have uploaded.

Do you mind check how do I generate the estimate for this? It seems that on my computer

TI().fit(ethanol) gives different result compared with TI_GQ().fit(ethanol)

hl2500: Thanks for the comment. For the ethanol (concat the coulomb and VDW) case, I think TI would take an extra difference between the last state of coulomb `(0.9531, 0.0)` and the first state of vdw `(1.0, 0.0254)`. I'm not sure if this is supposed to happen, as 1.0 is not part of the gaussian quadrature points. While if we use TI_GQ, the coulomb and vdw are calculated separately.
I also tried TI().fit(ethanol_Coulomb) and TI.fit(ethanol_VDW), separately. The results between TI and TI_GQ are more different on ethanol_Coulomb data than on ethanol_VDW data. This might because the quadrature points I used for the simulations are only 5 `(0.0469, 0.2308, 0.5000, 0.7692, 0.9531)` for Coulomb vs 7 `(0.0254, 0.1292, 0.2971, 0.5000, 0.7029, 0.8708, 0.9746)` for VDW. The less lambdas we use, the further away we get from the end points, for example, 0.0254 is closer to 0.0 than 0.0469, and 0.9746 is closer to 1.0 than 0.9531. If we use TI().fit(), usually simulations at 0.0 and 1.0 are expected or extrapolation might be needed, so estimation with TI() on this dataset might cause some inaccuracy and closer to 0.0 and 1.0 is more accurate. This is a difference between TI and TI_GQ. TI_GQ do not need simulations on 0.0 and 1.0.

xiki-tempula: I have generate the same ethanol in water test with the lambds of

```
      coul     vdw
0   0.0000  0.0000
1   0.0469  0.0000
2   0.2308  0.0000
3   0.5000  0.0000
4   0.7692  0.0000
5   0.9531  0.0000
6   1.0000  0.0000
7   1.0000  0.0254
8   1.0000  0.1292
9   1.0000  0.2971
10  1.0000  0.5000
11  1.0000  0.7029
12  1.0000  0.8708
13  1.0000  0.9746
14  1.0000  1.0000
```

I think the ideal case would be that if you run TI and TI_GQ on this set, you would get the same result.
[Archive.zip](https://github.com/alchemistry/alchemlyb/files/11423369/Archive.zip)

hl2500: Thanks for the files! Yes, that's correct. TI would consider all lambda points, while TI_GQ only counts for those between 0 and 1 if they are in the special_points list, but they have different weights for each lambda. The results should be similar between them when there are enough lambda points. I saw a ~0.2 kcal/mol difference in this case between TI and TI_GQ. My guess is that if we increase the number of lambda points the difference would be smaller. I'm trying 12 lambdas for coul and vdw, respectively, but I didn't include the (0.0000  0.0000), (1.0000  0.0000), and (1.0000  1.0000)

hl2500: @xiki-tempula I'm attaching the new dataset of ethanol in water. 12 gaussian quadrature points for coulomb and vdw respectively, plus 3 end states (0.0, 0.0), (1.0, 0.0), (1.0, 1.0). Could you please try it out when you get a chance?

The differences between TI and TI_GQ is ~0.058 kcal/mol in energy and ~0.00039 kcal/mol in error. The energy estimated by MBAR is slightly closer to TI_GQ. I guess more lambda points would bring the results from different methods closer to each other.

[ethanol.zip](https://github.com/alchemistry/alchemlyb/files/11445506/ethanol.zip)

xiki-tempula: @hl2500 Thank you very much for the waiting. I have tried your [ethanol.zip](https://github.com/alchemistry/alchemlyb/files/11445506/ethanol.zip) and I have two comments.

1. > I kind of want all estimator to retain the same interface. I wonder if it is possible to have lambda 0 and lambda 1 in `delta_f_` and `d_delta_f_`. The lambda 0 shall be quite simple which shall be zero. The lambda 1 could just be the same as the last value. Then add a note to the estimator saying that lambda 1 is just the same as the last lambda window.

2. I'm a bit confused as to the result that I got from this ethanol dataset. I tried
3.

hl2500: Thanks for the comments. Unlike the TI method, the values in the table are accumulative sum from one state (row: i) to another (column: j) :
```math
\sum_{k=i}^j \left(wight_k * \left< dHdl  \right>_k \right)
```
rather than the difference between the two states, so the diagonal values are not zeros and lower triangular values are zeros. I thought about putting the values with the opposite sign from the upper triangle, but they are not meaningful values. Or could I make the table symmetrical (the lower triangle is the same as the upper one)?

xiki-tempula: @hl2500 Thanks for the explanation. 
So in all other estimators (TI, MBAR, BAR), the values in the table are accumulative sum from one state (row: i) to another (column: j), so I kind of want to have a similar interface across all estimators.

> I thought about putting the values with the opposite sign from the upper triangle, but they are not meaningful values.

I do agree that it is not a meaningful value but I think putting the values with the opposite sign from the upper triangle might make it more similar to the other estimators.

hl2500: @xiki-tempula Thank you for the suggestions! 

I have entered the values into the lower triangle.

I'm wondering if we could keep the lambdas without adding the 0 and 1? If the simulations were not performed at 0 and 1, the lambdas 0 and 1 would not show in the tables from other estimators as well. For multi-lambda case, we also need to insert intermediate states, such as (1.0, 0.0) or (1.0, 1.0, 0.0). This is a little complicated as the number of lambda types and the number of values for each lambda type vary from different simulations.

Also, do you think I need to upload this ethanol test to alchemtest to replace the previous one with fewer lamndas? Could I change the name of this estimator to TIGQ?

Thank you for your time into this!

hl2500: @xiki-tempula Thank you for the suggestions. I have added the two end states and the new ethanol test has been uploaded, but it seemed that the checks have not been initiated yet.

xiki-tempula: Thanks for addressing the comments. The test is running now.

docs/estimators-ti.rst
```diff
@@ -45,15 +46,15 @@ The free energy differences (in units of :math:`k_B T`) between each :math:`\lam
 
 So we can get the endpoint differences (free energy difference between :math:`\lambda = 0` and :math:`\lambda = 1`) of each with::
 
-    >>> ti_coul.delta_f_.loc[0.00, 1.00]
+    >>> ti_coul.delta_f_.iloc[0, -1]
```
xiki-tempula, line 49: Now that we have .loc[0.00, 1.00], might be good to change this back.

docs/estimators-ti.rst
```diff
@@ -66,6 +67,12 @@ In addition to the free energy differences, we also have access to the errors on
     0.75  0.015038  0.011486  0.007458  0.000000  0.006447
     1.00  0.016362  0.013172  0.009858  0.006447  0.000000
 
+If the simulations are performed at certain gaussian quadrature points, :class:`~alchemlyb.estimators.TI_GQ` can be used to estimate the free energy.
```
xiki-tempula, line 70: Add a secondary title before this to highlight this part.

docs/estimators-ti.rst
```diff
@@ -66,6 +67,12 @@ In addition to the free energy differences, we also have access to the errors on
...
+The usage of :class:`~alchemlyb.estimators.TI_GQ` is similar to :class:`~alchemlyb.estimators.TI`, but instead of the free energy differences between 
+each :math:`\lambda` window, the values in the ``delta_f_`` and ``d_delta_f_`` tables are cumulative addition of estimation from one :math:`\lambda` window to another.
+To be consistent with :class:`~alchemlyb.estimators.TI` and other estimators, the diagonal values are set to zeros and two end states at :math:`\lambda`
+0 and 1 are added, although the simulation may not be performed at :math:`\lambda` 0 and 1. However, they are not considered to estimate the final results.
+For example, the values at :math:`\lambda` 0 is 0 and :math:`\lambda` 1 is the same as the previous gaussian quadrature point.
```
xiki-tempula, line 75: What do you mean `However, they are not considered to estimate the final results.`, I was under the impression that the for lambda points `[0.1127, 0.5, 0.88729]`, the result at point `0.88729` is the free energy difference between 0 and 1?

`:math:`\lambda` 0 is 0` is not the same as the previous gaussian quadrature point`

docs/estimators/alchemlyb.estimators.TI_GQ.rst
```diff
@@ -0,0 +1,12 @@
+.. _estimatators_TI_GQ:
+
+TI_GQ
+=====
+The :class:`~alchemlyb.estimators.TI_GQ` estimator is a simple implementation of `thermodynamic integration <https://en.wikipedia.org/wiki/Thermodynamic_integration>`_ that uses the `gaussian quadrature <https://en.wikipedia.org/wiki/Gaussian_quadrature>`_ for integrating the space between :math:`\left<\frac{dH}{d\lambda}\right>` values for each :math:`\lambda` sampled.
+To use this method, make sure that the simulations are performed at certain :math:`\lambda` values using fixed gaussian quadrature points (e.g., [He2020]_). Currently, upto 16 guassian quadrature points are supported. Otherwise, there would be associated errors.
```
xiki-tempula, line 6: up to

docs/estimators/alchemlyb.estimators.TI_GQ.rst
```diff
@@ -0,0 +1,12 @@
+.. _estimatators_TI_GQ:
+
+TI_GQ
+=====
+The :class:`~alchemlyb.estimators.TI_GQ` estimator is a simple implementation of `thermodynamic integration <https://en.wikipedia.org/wiki/Thermodynamic_integration>`_ that uses the `gaussian quadrature <https://en.wikipedia.org/wiki/Gaussian_quadrature>`_ for integrating the space between :math:`\left<\frac{dH}{d\lambda}\right>` values for each :math:`\lambda` sampled.
```
xiki-tempula, line 5: I don't think you need the word `simple`. Using the word simple imply that there is a more complex implementation and we are not doing that.

docs/estimators/alchemlyb.estimators.TI_GQ.rst
xiki-tempula: No need for `Otherwise, there would be associated errors.`

xiki-tempula: Might be good to add a rst table here to show the exact lambdas so people can check the doc and know the lambda that are needed.

src/alchemlyb/estimators/ti_gaussian_quadrature_.py
```diff
@@ -0,0 +1,231 @@
...
+            variance_list.append(new_variances)
+        
+        # add two end states at all lambda zeros and ones
+        if len(l_types) == 1:
+            index_list = [0.0] + index_list + [1.0]
+        else:
+            index_list = [tuple([0.0]*len(l_types))] + index_list + [tuple([1.0]*len(l_types))]
+
+        return lambda_list, dhdl_list, variance_list, index_list
+
```
xiki-tempula, line 231: tailing empty line

xiki-tempula (changes requested):
Thank you very much for continuously addressing my comments. The code part looks good to me now. We only need to polish the doc a bit and this PR should be good.

xiki-tempula:
@orbeckst Do you mind having a look at this PR as well? I think the code part is good and we just need to polish the doc a bit.

docs/estimators-ti.rst
hl2500: > What do you mean `However, they are not considered to estimate the final results.`

Sorry for the confusion. I meant that they didn't contribute to the final free energy values, as there were no estimations at lambda 0 and 1. I have removed this sentence. The next sentence should explain this.

xiki-tempula (approved):
Thanks very much for all the effort. I think I'm quite happy with this PR now.
However, apologise that we cannot merge this in now as it is a significant and important addition. So I want to make sure that no one would have strong opposition to this PR.
I purpose that we leave the PR here for one month and if no one is feeling strongly about it, I will merge this in. What do you think @orbeckst ?

hl2500:
> Thanks very much for all the effort. I think I'm quite happy with this PR now. However, apologise that we cannot merge this in now as it is a significant and important addition. So I want to make sure that no one would have strong opposition to this PR. I purpose that we leave the PR here for one month and if no one is feeling strongly about it, I will merge this in (aim for 2.2.0). What do you think @orbeckst ?

Okay, thank you so much for your guidance and time into this.

.github/workflows/ci.yaml
```diff
@@ -53,6 +53,10 @@ jobs:
         extra-specs: |
             python=${{ matrix.python-version }}
 
+    - name: Install alchemtest
```
xiki-tempula, line 56: Ok, I have merged the PR to the alchemtest. Please remove this.

xiki-tempula (changes requested):
remove the reference to the alchemtest PR and I will merge this PR next weekend. Thanks for waiting.

hl2500:
@xiki-tempula This part has been removed. Thank you!

xiki-tempula:
I have merge this PR thanks for all the contribution and the patience.

hl2500:
@xiki-tempula Thank you again for your guidance in this work!

Issue #309, opened 2023-04-07
Make pip test work in M1 Mac


xiki-tempula:
I'm preparing the 2.0.1 release, I tag it as 2.0.1 and upload it to `test.pypi.org`. Then I pip install the package locally with
`pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple alchemlyb=="2.0.1"`
Then I run the test with `pytest --pyargs alchemlyb`

However, the test would stuck indefinitely at `tests/test_workflow_ABFE.py`, where it seems to be stuck at
```
  File "/Users/zhiyiwu/miniconda3_x86/envs/test/lib/python3.11/site-packages/alchemlyb/tests/test_workflow_ABFE.py", line 39, in test_nofilematch
    ABFE(
  File "/Users/zhiyiwu/miniconda3_x86/envs/test/lib/python3.11/site-packages/alchemlyb/workflows/abfe.py", line 85, in __init__
    self.file_list = list(glob(reg_exp, recursive=True))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zhiyiwu/miniconda3_x86/envs/test/lib/python3.11/glob.py", line 28, in glob
    return list(iglob(pathname, root_dir=root_dir, dir_fd=dir_fd, recursive=recursive,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zhiyiwu/miniconda3_x86/envs/test/lib/python3.11/glob.py", line 97, in _iglob
    for name in glob_in_dir(_join(root_dir, dirname), basename, dir_fd, dironly,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zhiyiwu/miniconda3_x86/envs/test/lib/python3.11/glob.py", line 106, in _glob1
    names = _listdir(dirname, dir_fd, dironly)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zhiyiwu/miniconda3_x86/envs/test/lib/python3.11/glob.py", line 177, in _listdir
    return list(it)
           ^^^^^^^^
[...]

PR #310, opened 2023-04-07
309 make pip test work in m1 mac

xiki-tempula:
Ok, I got it to work now by changing `glob.glob` to `Path.glob`, these two should be the same but `Path.glob` works fine.
I tried 
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple alchemlyb=="2.0.1"
and then
pytest --pyargs alchemlyb

Where all tests passed fine.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -21,8 +21,8 @@ Fixes
   - Fix the dE method in u_nk2series to use the difference between two
     lambda columns instead of using the next lambda column or the previous
     column for the last window (issue #299, PR #300).
-  - Use Path.glob instead of glob.glob in ABFE workflow to get the files
-    (issue #309, PR #310).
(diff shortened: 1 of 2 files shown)


orbeckst (approved):
Did you check that this was the only use of `glob.glob()`? Any idea why this needs to be changed — it's odd that it would fail on M1?!?

In any case, it's a sensible change, so lgtm (just edited the CHANGES message).

orbeckst:
please squash-merge yourself when ready, @xiki-tempula

xiki-tempula:
@orbeckst Sorry I guess I haven't tried other machines. I haven't tried other machines but it hangs on my M1 Mac. As is said in #309. If I pip install the source it works fine. It only hangs when I upload it to pypi and download it to install.

PR #312, opened 2023-04-17
Change the documentation to make the behaviour of dir in ABFE pipeline more clear

### LINKED ISSUE #311: the dir keyword in ABFE pipeline no longer accepts wildcard
In the 2.0.1 #310 , I changed the way files are globbed from `glob.glob(dir + "/**/" + prefix + "*" + suffix)` into `Path(dir).glob("**/" + prefix + "*" + suffix)`. This means that the user can no longer supply wildcard into the `dir` argument. 

The documentation is not super clear on this.
```
dir ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Directory in which data files are stored. Default: os.path.curdir. The input files are searched using the pattern of dir + '/**/' + prefix + '*' + suffix.

prefix ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Prefix for datafile sets. Default: ‘dhdl’.

suffix ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Suffix for datafile sets. Default: ‘xvg’.
```

This is in some sense a good change as we expect the user to supply regular expression into `prefix` and in some sense `suffix` but we do expect the user to supply a real directory into `dir`.

xiki-tempula:
fix #311

### DIFF
--- a/src/alchemlyb/workflows/abfe.py
+++ b/src/alchemlyb/workflows/abfe.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import warnings
 from os.path import join
 from pathlib import Path
 
@@ -42,10 +43,10 @@ class ABFE(WorkflowBase):
         This option chooses the appropriate parser for the input file.
     dir : str
         Directory in which data files are stored. Default: os.path.curdir.
-        The input files are searched using the pattern of
-        ``dir + '/**/' + prefix + '*' + suffix``.
     prefix : str
-        Prefix for datafile sets. Default: 'dhdl'.
+        Prefix for datafile sets. This argument accepts regular expressions and
+        the input files are searched using
+        ``Path(dir).glob("**/" + prefix + "*" + suffix)``. Default: 'dhdl'.
     suffix : str
         Suffix for datafile sets. Default: 'xvg'.
     outdirectory : str
@@ -61,6 +62,8 @@ class ABFE(WorkflowBase):
 
 
     .. versionadded:: 1.0.0
+    .. versionchanged:: 2.0.1
+        The `dir` argument expect a real directory and wildcard are no longer accepted.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


src/alchemlyb/workflows/abfe.py
```diff
@@ -82,6 +85,10 @@ def __init__(
             f"{software}"
         )
         reg_exp = "**/" + prefix + "*" + suffix
+        if '*' in dir:
+            warnings.warn(f"A real directory is expected in `dir`={dir}, wildcard expressions should be supplied to `prefex`.")
```
orbeckst, line 89:
```suggestion
            warnings.warn(f"A real directory is expected in `dir`={dir}, wildcard expressions should be supplied to `prefix`.")
```

src/alchemlyb/workflows/abfe.py
```diff
@@ -61,6 +62,8 @@ class ABFE(WorkflowBase):
 
 
     .. versionadded:: 1.0.0
+    .. versionchanged:: 2.0.1
+        The `dir` argument expect a real directory and wildcard are no longer accepted.
```
orbeckst, line 66: We should not have done this API change in 2.0.1 as this breaks semver. 

However, the change is small enough so I am ok with just moving forward.

src/alchemlyb/workflows/abfe.py
orbeckst: If we find a `*`, will we then fail with ValueError on the next line?

src/alchemlyb/workflows/abfe.py
orbeckst, line 66:
```suggestion
        The `dir` argument expects a real directory without wildcards and wildcards will no longer
        work as expected. Use `prefix` to specify wildcard-based patterns to search under `dir`.  
```

orbeckst (approved):
That's a bit awkward because we shouldn't have merged PR #310 as a 2.0.1 as it technically breaks semantic versioning. However, as we have it released, this PR looks good.

I am just including two doc improvements.

src/alchemlyb/workflows/abfe.py
xiki-tempula: Depends on whether this is a directory or not if there is a directory that is literally named as `folder_*`, then it will be fine.

src/alchemlyb/workflows/abfe.py
xiki-tempula: Sorry, I hadn't thought of this before I do this change.

Issue #316, opened 2023-05-23
An engine agonistic parser


xiki-tempula:
I'm thinking of an engine agonistic IO that would load a dataframe directly from a file and save it to a file and read it back.
I think the pyarrow implementation of the [parquet](https://pandas.pydata.org/docs/dev/user_guide/io.html#io-parquet) is a good choice where it faithfully reproduce the complex index that we are using.

The interface that i'm thinking of would be

```
from alchemlyb.parsing.parquet import extract_dHdl, extract_u_nk
import pandas as pd

u_nk.to_parquet(path='u_nk.parquet', index=True)
dHdl.to_parquet(path='dHdl.parquet', index=True)

new_u_nk = extract_u_nk('u_nk.parquet', T=300)
new_dHdl = extract_dHdl('dHdl.parquet', T=300)

assert new_u_nk==u_nk
assert new_dHdl==dHdl
```

PR #317, opened 2023-05-23
parquet parser for dataframe serialisation

xiki-tempula:
This PR add a parquet parser, which will load the serialised dataframe and permits the usage of

```
    from alchemlyb.parsing.parquet import extract_dHdl, extract_u_nk
    import pandas as pd

    u_nk.to_parquet(path='u_nk.parquet', index=True)
    dHdl.to_parquet(path='dHdl.parquet', index=True)

    new_u_nk = extract_u_nk('u_nk.parquet', T=300)
    new_dHdl = extract_dHdl('dHdl.parquet', T=300) 
```
Fix https://github.com/alchemistry/alchemlyb/issues/316

### DIFF
--- a/docs/parsing.rst
+++ b/docs/parsing.rst
@@ -49,6 +49,21 @@ requires some care due to shortcomings in how pandas currently handles
 metadata (see issue `pandas-dev/pandas#28283 <https://github.com/pandas-dev/pandas/issues/28283>`_).
 
 
+Serialisation
+'''''''''''''
+
+To facilitate IO of the dataframe, the ``dHdl`` and ``u_nk`` could be serialised
+and load back via the `parquet <https://pandas.pydata.org/docs/user_guide/io.html#io-parquet>`_ ::
+
+    from alchemlyb.parsing.parquet import extract_dHdl, extract_u_nk
+    import pandas as pd
+
+    u_nk.to_parquet(path='u_nk.parquet', index=True)
+    dHdl.to_parquet(path='dHdl.parquet', index=True)
+
+    new_u_nk = extract_u_nk('u_nk.parquet', T=300)
+    new_dHdl = extract_dHdl('dHdl.parquet', T=300)
+
 
 .. _dHdl:
 
@@ -211,4 +226,5 @@ See the documentation for the package you are using for more details on parser u
     amber
     namd
     gomc
+    parquet
     
--- a/docs/parsing.rst
+++ b/docs/parsing.rst
@@ -52,8 +52,16 @@ metadata (see issue `pandas-dev/pandas#28283 <https://github.com/pandas-dev/pand
 Serialisation
 '''''''''''''
 
-To facilitate IO of the dataframe, the ``dHdl`` and ``u_nk`` could be serialised
-and load back via the `parquet <https://pandas.pydata.org/docs/user_guide/io.html#io-parquet>`_ ::
+Alchemlyb data structures (``dHdl`` and ``u_nk``) can be serialized as dataframes
+and made persistent.
+We use the `parquet <https://pandas.pydata.org/docs/user_guide/io.html#io-parquet>`_
+format for serializing (writing) to a file and de-serializing (reading) from a 
+parquet file.
+
+For serialization we simply use the :meth:`pandas.DataFrame.to_parquet` method of
+a :class:`pandas.DataFrame`. For loading alchemlyb data we provide the 
+:func:`alchemlyb.parsing.parquet.extract_dHdl` and 
+:func:`alchemlyb.parsing.parquet.extract_u_nk` functions as shown in the example::
 
     from alchemlyb.parsing.parquet import extract_dHdl, extract_u_nk
     import pandas as pd
--- a/src/alchemlyb/parsing/parquet.py
+++ b/src/alchemlyb/parsing/parquet.py
@@ -9,7 +9,7 @@ def extract_u_nk(path, T):
 
     The parquet file should be serialised from the dataframe output
     from any parser with command
-    (`u_nk_df.to_parquet(path=path, index=True)`).
+    (``u_nk_df.to_parquet(path=path, index=True)``).
 
     Parameters
     ----------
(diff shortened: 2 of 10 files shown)


xiki-tempula:
@orbeckst Do you mind have a review of this PR, please? Thank you.

docs/parsing.rst
```diff
@@ -49,6 +49,21 @@ requires some care due to shortcomings in how pandas currently handles
 metadata (see issue `pandas-dev/pandas#28283 <https://github.com/pandas-dev/pandas/issues/28283>`_).
 
 
+Serialisation
+'''''''''''''
+
+To facilitate IO of the dataframe, the ``dHdl`` and ``u_nk`` could be serialised
+and load back via the `parquet <https://pandas.pydata.org/docs/user_guide/io.html#io-parquet>`_ ::
+
```
orbeckst, line 57:
```suggestion
Alchemlyb data structures (``dHdl`` and ``u_nk``) can be serialized as dataframes
and made persistent.
We use the `parquet <https://pandas.pydata.org/docs/user_guide/io.html#io-parquet>`_
format for serializing (writing) to a file and de-serializing (reading) from a 
parquet file.

For serialization we simply use the :meth:`pandas.DataFrame.to_parquet` method of
a :class:`pandas.DataFrame`. For loading alchemlyb data we provide the 
:func:`alchemlyb.parsing.parquet.extract_dHdl` and 
:func:`alchemlyb.parsing.parquet.extract_u_nk` functions as shown in the example::

```

src/alchemlyb/parsing/parquet.py
```diff
@@ -0,0 +1,77 @@
...
+from . import _init_attrs
+
+
+@_init_attrs
+def extract_u_nk(path, T):
+    r"""Return reduced potentials `u_nk` from a pandas parquet file.
+
+    The parquet file should be serialised from the dataframe output
+    from any parser with command
+    (`u_nk_df.to_parquet(path=path, index=True)`).
```
orbeckst, line 12: Need double back-ticks for proper code markup.
```suggestion
    (``u_nk_df.to_parquet(path=path, index=True)``).
```

src/alchemlyb/parsing/parquet.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Parameters
+    ----------
+    path : str
+        Path to parquet file to extract dataframe from.
+    T : float
+        Temperature in Kelvin the simulations sampled.
```
orbeckst, line 19:
```suggestion
        Temperature in Kelvin of the simulations.
```

src/alchemlyb/parsing/parquet.py
```diff
@@ -0,0 +1,77 @@
...
+    ----
+    pyarraw serializers would handle the float or string column name fine but will
+    convert multi-lambda column name from `(0.0, 0.0)` to `"('0.0', '0.0')"`.
+    This parser will restore the correct column name.
+
+
+    .. versionadded:: 2.1.0
+
+    """
+    u_nk = extract_dHdl(path, T)
```
orbeckst, line 37: confusing, just be explicit
```suggestion
    u_nk = pd.read_parquet(path)
```

src/alchemlyb/parsing/parquet.py
orbeckst: State that T is ignored but included for API compatibility

src/alchemlyb/parsing/parquet.py
```diff
@@ -0,0 +1,77 @@
...
+    The parquet file should be serialised from the dataframe output
+    from any parser with command
+    (`dHdl_df.to_parquet(path=path, index=True)`).
+
+    Parameters
+    ----------
+    path : str
+        Path to parquet file to extract dataframe from.
+    T : float
+        Temperature in Kelvin the simulations sampled.
```
orbeckst, line 66: State that T is ignored but included for API compatibility

orbeckst (changes requested):
Basic functionality looks good but some request for changes
- docs and clearer code (see comments)
- update all build files with new pyarrow dependency: [environment.yml](https://github.com/alchemistry/alchemlyb/blob/master/environment.yml), [setup.py](https://github.com/alchemistry/alchemlyb/blob/master/setup.py), + raise issue in https://github.com/conda-forge/alchemlyb-feedstock/ to add pyarrow to the deps of the [meta.yml](https://github.com/conda-forge/alchemlyb-feedstock/blob/main/recipe/meta.yaml) of the conda-forge package (and link to this issue)

src/alchemlyb/parsing/parquet.py
xiki-tempula: Sorry but parquet serialisation doesn't preserve the df.attrs. So the dataframe loaded here doesn't contain temperature and the temperature is assigned via this `T`.

xiki-tempula:
@orbeckst Thanks for the review. I have change the [environment.yml](https://github.com/alchemistry/alchemlyb/blob/master/environment.yml). I will raise an issue with [meta.yml](https://github.com/conda-forge/alchemlyb-feedstock/blob/main/recipe/meta.yaml) when we are doing a release.

So the T problem is a bit tricky in that parquet serialisation doesn't preserve the df.attrs. So the dataframe loaded here doesn't contain temperature and the temperature is assigned via extract function. I will add a not to state that.

orbeckst:
> So the T problem is a bit tricky in that parquet serialisation doesn't preserve the df.attrs. So the dataframe loaded here doesn't contain temperature and the temperature is assigned via extract function. I will add a not to state that.

Ok, I blithely assumed that parquet would save _everything_ — then `T` is not optional. Definitely document this requirement.

xiki-tempula:
@orbeckst 

> Ok, I blithely assumed that parquet would save everything

I'm annoyed by this as well but it seems that parquet is still the best serialiser. This is the only format that preserves index besides to_pickle. Also it faithfully preserve all the data in its original datatype. The to_pickle, though it preserves everything, is slow to read and write and also would not be safe between different versions of pandas.

> Definitely document this requirement.

I have added this as a note to both alchemlyb.parsing.parquet.extract_u_nk and alchemlyb.parsing.parquet.extract_dHdl.

orbeckst (approved):
lgtm

orbeckst:
Please squash-merge when ready.

Issue #320, opened 2023-06-01
enable bootstrap


xiki-tempula:
Given that pymbar 4 enables one to use bootstrap to compute uncertainty, we shall enable it as well.

PR #322, opened 2023-06-03
Enable MBAR to do bootstrap error estimation

xiki-tempula:
Implement #320


src/alchemlyb/workflows/abfe.py
```diff
@@ -408,6 +408,9 @@ def estimate(self, estimators=("MBAR", "BAR", "TI"), **kwargs):
             'MBAR']. Note that the estimators are in their original form where
             no unit conversion has been attempted.
 
+        .. versionchanged:: 2.1.0
+        `n_bootstraps=50` is used for MBAR estimator.
```
orbeckst, line 412:
```suggestion
           `n_bootstraps=50` is used for MBAR estimator.
```

src/alchemlyb/workflows/abfe.py
```diff
@@ -441,7 +444,12 @@ def estimate(self, estimators=("MBAR", "BAR", "TI"), **kwargs):
         for estimator in estimators:
             if estimator == "MBAR":
                 self.logger.info("Run MBAR estimator.")
-                self.estimator[estimator] = MBAR(**kwargs).fit(u_nk)
+                self.logger.info(
+                    "By default, n_bootstraps=50 is used to estimate the "
+                    "MBAR error. Supply n_bootstraps=0 to use analytic error."
+                )
+                estimator_kwargs = {"n_bootstraps": 50, **kwargs}
```
orbeckst, line 451: This will fail when I actually supply `n_bootstraps` as a kwarg because then the dict contains the same key twice ... a test would have caught it.

Instead, I would do
```python
# use 0 as default to remain backwards-compatible
# change in release 2.3
kwargs.setdefault('n_bootstraps', 0)
...
...
if estimator == "MBAR":
  estimator_kwargs = kwargs
  self.logger.info("Run MBAR estimator")
  self.logger.info("Estimating MBAR error with n_bootstraps={n_bootstraps} (0 is analytical estimate)", estimator_kwargs)
elif estimator == "BAR":
  estimator_kwargs = kwargs.copy()
  estimator_kwargs.pop('n_bootstraps', None)
  ...
  self.estimator[estimator] = BAR(**kwargs).fit(u_nk)
```
i.e., modify the copy of kwargs as needed but keep everything in kwargs.

src/alchemlyb/workflows/abfe.py
```diff
@@ -441,7 +444,12 @@ def estimate(self, estimators=("MBAR", "BAR", "TI"), **kwargs):
         for estimator in estimators:
             if estimator == "MBAR":
                 self.logger.info("Run MBAR estimator.")
-                self.estimator[estimator] = MBAR(**kwargs).fit(u_nk)
+                self.logger.info(
+                    "By default, n_bootstraps=50 is used to estimate the "
+                    "MBAR error. Supply n_bootstraps=0 to use analytic error."
```
orbeckst, line 449: We should not change the default without a deprecation. I'd issue a DeprecationWarning to state that the error estimate will change in release 2.3 to n_bootstraps=50. 

Until then, keep 
```python
kwargs.setdefault('n_bootstraps', 0)
```
to keep the old behavior.

orbeckst: I'd also use a more informative status message (see other comment) that reports on n_bootstrap.

Use a DeprecationWarning for letting users know about coming changes.

CHANGES
```diff
@@ -20,6 +20,7 @@ The rules for this file:
 Enhancements
   - Add a parser to read serialised pandas dataframe (parquet) (issue #316, PR#317).
   - workflow.ABFE allow parquet as input (issue #316, PR#317).
+  - Allow MBAR estimator to use bootstrap to compute error (issue #320, PR#322).
```
orbeckst, line 24: Add a deprecation for using analytical error estimate as the default in ABFE, will change to using 50 bootstraps in 2.3 (see other comments)

orbeckst (changes requested):
Overall looking good, by under semver we can't just change the way that ABFE computes the error. We need to **deprecate** the default (analytical) and then switch over to `n_bootstraps=50` in, say, release 2.3.

The kwargs handling of n_bootstraps needs improving (and possibly a test — sorry).

src/alchemlyb/workflows/abfe.py
xiki-tempula: Sorry, what do you mean this will fail?

```
>>> kwargs = {"n_bootstraps": 0}
>>> estimator_kwargs = {"n_bootstraps": 50, **kwargs}
>>> print(estimator_kwargs)
{'n_bootstraps': 0}
```

Gives the expected behaviour where kwargs overrides the default in estimator_kwargs?

orbeckst: You're right, I am a bit surprised. (I was in the process of trying it out but then wrote the review comment without finishing my testing.)

My concern remains that the default ought to be n_bootstraps=0 to initially have the same behavior.

src/alchemlyb/workflows/abfe.py
xiki-tempula: Yes, I think for this PR I will just add a DeprecationWarning without changing the behaviour.

CHANGES
```diff
@@ -23,10 +23,15 @@ Changes
...
 
 Fixes
   - Fix the case where visualisation.plot_convergence would fail when the final
-   error is NaN (issue #318, PR#319).
+    error is NaN (issue #318, PR#319).
+
+DeprecationWarning
+  - The default MBAR error estimator in workflow.ABFE.estimate will change from
+  analytic to bootstrap=50 (issue #320, PR#322).
```
orbeckst, line 34: indentation

orbeckst: say when it will change (target release, current + 2 = 2.3 I think)

src/alchemlyb/workflows/abfe.py
```diff
@@ -403,6 +403,9 @@ def estimate(self, estimators=("MBAR", "BAR", "TI"), **kwargs):
             'MBAR']. Note that the estimators are in their original form where
             no unit conversion has been attempted.
 
+        .. versionchanged:: 2.1.0
+        DeprecationWarning for using analytic error for MBAR estimator.
```
orbeckst, line 407: indent

CHANGES
orbeckst: Or 2.2, if you want to be aggressive, but state the target release.

src/alchemlyb/estimators/mbar_.py
```diff
@@ -67,13 +73,15 @@ def __init__(
         relative_tolerance=1.0e-7,
         initial_f_k=None,
         method="robust",
+        n_bootstraps=0,
```
orbeckst, line 76: I'd add a reminder comment
```python
 n_bootstraps=0,    # release 2.2: change to 50 (see PR #322)
```

orbeckst (changes requested):
minor things

src/alchemlyb/estimators/mbar_.py
xiki-tempula: So my feeling is that for the MBAR estimator, we retain the current behaviour unless people manually turn it on as bootstrap does come with a computational cost.

It is only for the estimate method in workflow.ABFE where I know that the input is decorrelated and I'm prepared to pay the computational cost, I will set the default to 50.

orbeckst: Yes, that's ok, we just can't make the default behave differently without a minimal warning period.

orbeckst (approved):

PR #325, opened 2023-06-11
The statistical inefficiency is logged at the debug level

### LINKED ISSUE #295: The output of the function `statistical_inefficiency` doesn't include statistical inefficiency value
The current `statistical_inefficiency` function only returns the subsampled data, but I also want the value of statistical inefficiency. And I have to use the statistical_inefficiency in 'pymbar', and do a similar calculation again.

It'd help if `statistical_inefficiency` returns the subsampled data as well as the value of statistical inefficiency or has an option to control whether to return the value.

xiki-tempula:
Fix #295

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,8 @@ Changes
   - Use loguru instead of logging for log (issue #301, PR #303).
 
 Enhancements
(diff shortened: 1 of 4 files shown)


orbeckst (approved):
Looks good.

Maybe I would log at INFO but we can always increase the level later.

PR #338, opened 2023-12-24
Fix 337

xiki-tempula:
- Warning issued when the series is `None` for `statistical_inefficiency`
  - ValueError issued when `df` and `series` for `statistical_inefficiency`
    doesn't have the same length


src/alchemlyb/preprocessing/subsampling.py
```diff
@@ -363,6 +363,15 @@ def _prepare_input(df, series, drop_duplicates, sort):
     series : Series
         Formatted Series.
     """
+    if series is None:
+        logger.warning(
```
orbeckst, line 367: I think we also should raise a real warning
```python
warnings.warn(...)
```
in addition to logging one.

src/alchemlyb/tests/test_preprocessing.py
```diff
@@ -544,3 +544,16 @@ def test_statistical_inefficiency(self, caplog, u_nk):
...
             assert "Number of uncorrelated samples:" in caplog.text
+
+
+def test_unequil_input(dHdl):
+    with pytest.raises(ValueError, match="should be same as the length of series"):
+        statistical_inefficiency(dHdl, series=dHdl[:10])
+
+
+def test_series_none(dHdl, caplog):
+    statistical_inefficiency(dHdl, series=None)
```
orbeckst, line 555: Also test with `pytest.warns()`, see above.

orbeckst (changes requested):
Thanks for addressing the issue. Could you please also raise a Python warning?

orbeckst (approved):
Warning instead of logger works for me (although I would also be happy with warning + logger if you prefer both). Thanks for addressing my comments!

orbeckst:
@xiki-tempula looks good as it is but if you want to add the logger back then please do. Please squash-merge whenever you're ready.

PR #340, opened 2023-12-30
Fix #331

### LINKED ISSUE #331: Parquet file can now save metadata
With this PR (https://github.com/pandas-dev/pandas/pull/54346), pandas can now save the metadata in parquet files as well. When a release of the new pandas (2.1) has been made. I need to change the doc as well as add a check to the extract function.

xiki-tempula:
Allow metadata to be loaded from the serialised parquet file.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,9 @@ The rules for this file:
 */*/2023 hl2500
 
   * 2.2.0
+
+Changes
+ - For pandas>=2.1, metadata will be loaded from the parquet file (issue #331, PR #326).
   
 Enhancements
   - Add a TI estimator using gaussian quadrature to calculate the free energy.
--- a/src/alchemlyb/parsing/parquet.py
+++ b/src/alchemlyb/parsing/parquet.py
@@ -1,9 +1,42 @@
 import pandas as pd
+from loguru import logger
 
 from . import _init_attrs
 
 
-@_init_attrs
+def _check_metadata(path: str, T: float) -> pd.DataFrame:
+    """
+    Check if the metadata is included in the Dataframe and has the correct
+    temperature.
+
+    Parameters
+    ----------
+    path : str
+        Path to parquet file to extract dataframe from.
+    T : float
+        Temperature in Kelvin of the simulations.
+
+    Returns
+    -------
+    DataFrame
+    """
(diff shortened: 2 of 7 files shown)


xiki-tempula:
@orbeckst The py3.8 is deprecated in #342 and the RTD is fixed in #338

CHANGES
```diff
@@ -17,6 +17,9 @@ The rules for this file:
 */*/2023 hl2500
 
   * 2.2.0
+
```
orbeckst, line 20: add your GH handle to the CHANGELOG

CHANGES
```diff
@@ -17,6 +17,9 @@ The rules for this file:
...
+Changes
```
orbeckst, line 21: state that the minimal supported version of pandas is now >= 2.1

src/alchemlyb/parsing/parquet.py
```diff
@@ -1,9 +1,42 @@
 import pandas as pd
+from loguru import logger
 
 from . import _init_attrs
 
 
-@_init_attrs
```
orbeckst, line 6: Is the decorator not needed anymore for `extract_u_nk()`?

src/alchemlyb/parsing/parquet.py
```diff
@@ -1,9 +1,42 @@
...
+def _check_metadata(path: str, T: float) -> pd.DataFrame:
```
orbeckst, line 7: The name is a bit misleading because it does not only check but also return a df. I would choose a name such as `_read_parquet_with_metadata` or similar.

orbeckst (changes requested):
Small changes/suggestions, otherwise looking sensible.

src/alchemlyb/parsing/parquet.py
xiki-tempula: Not anymore as the attrs are loaded from the parquet file

orbeckst: thanks for the explanation

orbeckst:
RTD fails https://readthedocs.org/projects/alchemlyb/builds/23006311/ because PR #338 is not merged yet. I'll merge nevertheless.

PR #342, opened 2023-12-30
Fix 341

xiki-tempula:
Deprecate python 3.8 add support for python 3.12


CHANGES
```diff
@@ -17,7 +17,10 @@ The rules for this file:
 */*/2023 hl2500
 
   * 2.2.0
-  
+
+Changes
+  - Deprecate python 3.8 add support for python 3.12 (issue #341, PR #304).
```
orbeckst, line 22: We are removing, not deprecating (which I understand to be keeping a feature but announcing that it will go away).
```suggestion
  - add support for Python 3.12, remove Python 3.8 support (issue #341, PR #304).
```

orbeckst (approved):
lgtm (will just adjust the CHANGELOG text)

PR #346, opened 2024-01-08
2.2.0 release

xiki-tempula:
I have uploaded to pip and test that it works fine.


orbeckst:
Why is py 3.12 installation failing?

orbeckst:
Is pymbar not packaged for 3.12 - https://github.com/alchemistry/alchemlyb/actions/runs/7451751133/job/20274729913?pr=346#step:3:49 ?

orbeckst (changes requested):
Currently fails Py 3.12 CI: apparently it cannot find pymbar.

We can't really release it until this is fixed.

xiki-tempula:
@orbeckst 
Ok, I guess we have to remove the support for python 3.12. I used to have this problem with python 3.12 before with `mamba-org/provision-with-micromamba@main`.
Then I had a look and found `mamba-org/provision-with-micromamba@main` to be deprecated, so I switch to `mamba-org/setup-micromamba@main`, the problem kind of goes away.
Then in this PR, I found that `mamba-org/setup-micromamba@main` tells me that `extra-specs` is deprecated and one should use `create-args` instead, so I switch to `create-args` and the python 3.12 problem goes back.

Now I had a more close look and it seems that there is no python 3.12 version for JAX, which is a dependency for pymbar.

The reason that `mamba-org/setup-micromamba@main` initially fixed the problem is because `extra-specs` is not a keyword for `mamba-org/setup-micromamba@main` so all of them are using python 3.11 instead of the specified python version.

orbeckst:
Are these failures related to versioneer? https://github.com/alchemistry/alchemlyb/actions/runs/8396562936/job/22998193364?pr=346#step:3:595

xiki-tempula:
@orbeckst Yes, I'm fixing that now. I think it might be good for alchemtest to adopt the MolSSI cookiecutter then we could merge those fixed from them easily.

orbeckst:
closing and reopening, now that alchemtest has been update https://github.com/alchemistry/alchemtest/pull/93

xiki-tempula:
@orbeckst I think we could do the release before the doc PR as the doc will not affect the release.
I have deployed to test pypi and tested it locally which is fine.

orbeckst:
Thank you @xiki-tempula ! Looking all good, go ahead 🚀 !

PR #348, opened 2024-03-13
Lammps Support

jaclark5:
**This is a Draft** I am actively using this and expect it's ready to go, but still needs tests.

This module will allow for alchemlyb to be used with LAMMPS outputs. An independent package for easily generating the LAMMPS input files will shortly follow, although the functions have been designed to be flexible.

In addition to the ``extract_dHdl`` and ``extract_u_nk`` functions, additional functions, ``extract_dHdl_from_u_n`` and ``extract_u_nk_from_u_n`` have been added to allow these data frames to be generated from potential energy information that is separable from the target variable (e.g., epsilon in the LJ).

PR #351, opened 2024-03-24
Fix #345

### LINKED ISSUE #345: Run the TestInit.test_nofilematch test in a tempdir
The current test is
```
    def test_nofilematch(self):
        with pytest.raises(ValueError, match="No file has been matched to"):
>           ABFE(
                dir="./",
                prefix="dhdl",
                suffix="xvg",
                T=310,
            )
```
Which is fine it's own sense but will try to search through all the subfolders in dir="./". So if you run it in your root path '~', it will try to access all the files which might annoy some security people. Once I had time should make a temp dir for this test.

xiki-tempula:
Fix #345


devtools/conda-envs/test_env.yaml
```diff
@@ -24,5 +24,5 @@ dependencies:
 - sphinx_rtd_theme
 
 - pip:
-  - alchemtest @ https://github.com/alchemistry/alchemtest/archive/master.zip
+  - alchemtest @ git+https://github.com/xiki-tempula/alchemtest.git@feat_main
```
orbeckst, line 27: change back

.github/workflows/ci.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
       with:
         environment-file: devtools/conda-envs/test_env.yaml
         environment-name: test
-        extra-specs: |
+        create-args: |
```
orbeckst, line 53: Is this also being fixed in #346 ?

orbeckst (changes requested):
LGTM, just fix the CI that pulls from your personal GH.

.github/workflows/ci.yaml
xiki-tempula: Yes.

xiki-tempula:
Remove the py3.12 support now. I will add the py3.12 support in another PR, where I think I will update the versioneer. I would hope to do the same some of updating to the cookie cutter here but I guess that will generate big merge conflict for https://github.com/alchemistry/alchemlyb/pull/348

docs/requirements.yaml
```diff
@@ -0,0 +1,9 @@
+name: docs
+channels:
+dependencies:
+    # Base depends
+  - python=3.10
```
xiki-tempula, line 5: Pin it to a python version so it would not break just due to some python update.

orbeckst (approved):

PR #354, opened 2024-03-31
Update versioneer

xiki-tempula:
Fix #352, builds on top of https://github.com/alchemistry/alchemlyb/pull/351

PR #355, opened 2024-03-31
Fix #350

### LINKED ISSUE #350: The data_fraction column in forward_backward_convergence will be scaled by the unit converter in ABFE workflow
In the ABFE workflow, the convergence dataframe is passed into the unit converter to ensure that it outputs the correct unit, which will convert `data_fraction` column as well. For example, if the output is kcal/mol, the `data_fraction` column which is for example `[0.0, 0.1, 0.2, ..., 1.0]` will be converted to `[0.061, 0.123, 0.184, ..., 0.616]`.

xiki-tempula:


orbeckst:
This branch also contains the versioneer updates from PR #354. Can you rebase or merge once that PR is merged to get a cleaner diff?

orbeckst (approved):
lgtm, thanks for the fix!

PR #357, opened 2024-05-15
Speed up the MBAR calculation

xiki-tempula:
I have adapted @mrshirts 's advice of using BAR results as initial guess into the MBAR, which seems to provide a 5-fold speed up. 
```
from alchemtest.gmx import load_ABFE
import pandas as pd
df_list = [extract_u_nk(data, 310) for data in load_ABFE()['data']['complex']]
df = pd.concat(df_list)
mbar = MBAR(initial_f_k=None)
%timeit mbar.fit(df)
1.66 s ± 7.42 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
mbar = MBAR(initial_f_k='BAR')
%timeit mbar.fit(df)
248 ms ± 2.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
I tried to optimise the convergence detection as well, where using the result from the previous MBAR run as input into the next MBAR run. Compared to using BAR as input, this approach still provides some speed up.

```
from alchemlyb.convergence import forward_backward_convergence
%timeit forward_backward_convergence(df_list)
```

Using BAR as initial guess
3.58 s ± 35.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
Using the previous step as initial guess.
2.25 s ± 23.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


src/alchemlyb/convergence/convergence.py
```diff
@@ -82,7 +82,7 @@ def forward_backward_convergence(df_list, estimator="MBAR", num=10, **kwargs):
         raise ValueError(msg)
     else:
         # select estimator class by name
-        estimator_fit = estimators_dispatch[estimator](**kwargs).fit
```
orbeckst, line 85: I'd rename `estimator_fit` to `my_estimator` — that will make `my_estimator.fit()` a lot nicer to read.

(Unfortunately, `estimator` is already a variable.)

.github/workflows/ci.yaml
```diff
@@ -64,9 +64,9 @@ jobs:
         MPLBACKEND: agg
 
     - name: Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
+        token: ${{ secrets.CODECOV }}
```
orbeckst, line 69: Is this a change due to a merge or did you smuggle it into this PR??

src/alchemlyb/convergence/convergence.py
```diff
@@ -94,7 +94,9 @@ def forward_backward_convergence(df_list, estimator="MBAR", num=10, **kwargs):
         for data in df_list:
             sample.append(data[: len(data) // num * i])
         sample = concat(sample)
-        result = estimator_fit(sample)
+        result = estimator_fit.fit(sample)
+        if estimator == "MBAR":
+            estimator_fit.initial_f_k = result.delta_f_.iloc[0, :]
```
orbeckst, line 99: I didn't fully follow the code path but how is `delta_f` already initialized with the BAR results?

src/alchemlyb/estimators/mbar_.py
```diff
@@ -1,7 +1,12 @@
+from __future__ import annotations
```
orbeckst, line 1: Which versions of python support annotations?

orbeckst (changes requested):
Neat idea and cool that it gives a real performance boost.

Definitely needs documentation!

.github/workflows/ci.yaml
xiki-tempula: So you could see that the CI failed for codecov token issues, I tried to see if I could fix it.

src/alchemlyb/estimators/mbar_.py
xiki-tempula: 3.10 supports full functionality of annotations. The old versions only supports less optimal way of annotating it.

src/alchemlyb/convergence/convergence.py
xiki-tempula: So you did a MBAR calculation first `result = estimator_fit.fit(sample)`
Which populates the `result.delta_f_` with for the current MBAR results.
This MBAR results is then propagated back as initial guess for the next calculation.

xiki-tempula:
Some codecov is still failing for token issue but otherwise this seems fine. When this PR is merged. Do you mind if I do a new 2.2.1 release? I think the speed increase is quite significant.

CHANGES
```diff
@@ -14,6 +14,14 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 
+**/**/2024 xiki-tempula
+
+  * 2.2.1
+
```
orbeckst, line 20: document default change
```suggestion

Changes
  - Default value for keyword argument `initial_nk` of the MBAR estimator was
    changed to "BAR" (run an initial BAR calculation before MBAR) instead of 
    `None` (start from all zeros) as this change provides a sizable speedup (PR #357)
    
```

CHANGES
```diff
@@ -14,6 +14,14 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 
+**/**/2024 xiki-tempula
+
+  * 2.2.1
```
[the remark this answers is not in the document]
orbeckst:
```suggestion
  * 2.3.0
```

src/alchemlyb/tests/test_fep_estimators.py
```diff
@@ -151,3 +151,10 @@ def test_bootstrap(gmx_benzene_Coulomb_u_nk):
 
     assert mbar_bootstrap_mean == mbar_mean
     assert mbar_bootstrap_err != mbar_err
+
```
orbeckst, line 154: add a test for the old `initial_nk = None` behavior

orbeckst (changes requested):
Let's make a 2.3.0 release (see my suggested changes — you can just accept the whole bunch) and document carefully API change (new default).

Please add a test for `MBAR(initial_nk=None)` for the old behavior, given that default changed.

src/alchemlyb/estimators/mbar_.py
xiki-tempula: This is an issue if we don't have `from __future__ import annotations`. Adding this allow py3.9 to be supported.

orbeckst (approved):

PR #358, opened 2024-05-17
Improve the error handling in the backward and forward convergence

xiki-tempula:
In the backward and forward convergence, for the initial set of points, which uses for example the first 10% of the date, it could be the cases where due to the fact that there are not many data points, so the overlap is pretty bad, which gives terrible statistical error.
This PR is build on https://github.com/alchemistry/alchemlyb/pull/357.
The PR allow the user to specify a error tolerance. If the statistical error is too bad, use the bootstrap error instead https://github.com/choderalab/pymbar/issues/519.


orbeckst:
@xiki-tempula you'll have to manually resolve these conflicts

CHANGES
```diff
@@ -24,6 +24,7 @@ Changes
     `None` (start from all zeros) as this change provides a sizable speedup (PR #357)
     
 Enhancements
+  - `forward_backward_convergence` uses the bootstrap error when the statistical error is too large. (PR #357)
```
orbeckst, line 27:
```suggestion
  - `forward_backward_convergence` uses the bootstrap error when the statistical error
    is too large. (PR #358)
```
fix PR reference

src/alchemlyb/convergence/convergence.py
```diff
@@ -35,6 +39,9 @@ def forward_backward_convergence(df_list, estimator="MBAR", num=10, **kwargs):
            Lower case input is also accepted until release 2.0.0.
     num : int
         The number of time points.
+    error_tol : float
+        The maximum error tolerated for analytic error. If the analytic error is
+        bigger than the error tolerance, the bootstrap error will be used.
```
orbeckst, line 44:
```suggestion
        bigger than the error tolerance, the bootstrap error will be used.

        .. versionadded:: 2.3.0

```

src/alchemlyb/convergence/convergence.py
```diff
@@ -122,23 +117,11 @@ def forward_backward_convergence(df_list, estimator="MBAR", num=10, **kwargs):
         sample = []
         for data in df_list:
             sample.append(data[-len(data) // num * i :])
-        sample = concat(sample)
-        result = my_estimator.fit(sample)
-        if estimator == "MBAR":
-            my_estimator.initial_f_k = result.delta_f_.iloc[0, :]
```
orbeckst, line 128: by deleting this block, aren't you undoing some of the changes from PR #357, namely the CHANGES entry
> forward_backward_convergence` uses the result from the previous step as the initial guess for the next step. (PR #357)

Please confirm that this is what you want to do, and if so, update CHANGES.

src/alchemlyb/convergence/convergence.py
```diff
@@ -93,23 +100,11 @@ def forward_backward_convergence(df_list, estimator="MBAR", num=10, **kwargs):
         sample = []
         for data in df_list:
             sample.append(data[: len(data) // num * i])
-        sample = concat(sample)
-        result = my_estimator.fit(sample)
-        if estimator == "MBAR":
-            my_estimator.initial_f_k = result.delta_f_.iloc[0, :]
```
orbeckst, line 99: by deleting this block, aren't you undoing some of the changes from PR #357, namely the CHANGES entry
> forward_backward_convergence` uses the result from the previous step as the initial guess for the next step. (PR #357)

Please confirm that this is what you want to do, and if so, update CHANGES.

orbeckst (changes requested):
Did you remove the speed-up for the convergence estimate (ie using the bar estimate)? If so update CHANGES (and docs where necessary).

src/alchemlyb/convergence/convergence.py
xiki-tempula: Sorry this is unintended change. I have added it back.

orbeckst: glad I caught it!

orbeckst (approved):

PR #328, opened 2023-07-08
JOSS paper

xiki-tempula:
@orbeckst I have made a quick draft of the JOSS paper.
I have only written the summary and the Statement of need and left other things as the template.
The guide line is here https://joss.readthedocs.io/en/latest/submitting.html#example-paper-and-bibliography

## Co-authors
Substantial contributors to alchemlyb are invited to become authors of the paper. See https://github.com/alchemistry/alchemlyb/pull/328#issuecomment-2123030068 below for details.

## Building the PDF
As part of the GitHub actions workflow, a full PDF version of the paper is built and you can download it as an artifact.

If you want to build the PDF locally you can use `pandoc`
```bash
pandoc --citeproc -o paper.pdf  paper.md
```

PR #370, opened 2024-05-28
adopt SPEC 0 instead of NEP 29

### LINKED ISSUE #369: switch from NEP29 to SPEC0
[NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) is superceded by [SPEC 0](https://scientific-python.org/specs/spec-0000/). We should adopt it officially as there isn't much (anything?) that really changes for us.

Mostly that's a doc change...

orbeckst:
- fix #369
- once this is fixed, we can update the JOSS paper draft PR #328 to only mention SPEC 0

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,14 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" https://semver.org
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


xiki-tempula (approved):
LGTM. Thanks.

PR #378, opened 2024-07-05
Prepare 2.3.1

xiki-tempula:
For some reason the `rst` readme cannot be used as long description, changed this to `markdown` file and it seems to work.
Build and tested on pypi testing channel


setup.py
```diff
@@ -41,7 +41,7 @@
     packages=find_packages("src"),
     package_dir={"": "src"},
     license="BSD",
-    long_description=open("README.rst").read(),
+    long_description=open("README.md").read(),
     long_description_content_type="text/x-rst",
```
orbeckst, line 45: Change the content description.

orbeckst (changes requested):
Not sure why rst isn't working, I normally run `twine check` and it tells me what's wrong. In any case, markdown is ok, too (even though I generally consider it a poor replacement for reST ... but README is fine).

However, you need to fix-up a whole bunch places.
- MANIFEST.in
- setup.py
- CHANGES:
  - add your name to 2.3.1
  - add the release date

Maybe other places, check with `git grep README`.

orbeckst (approved):

Issue #380, opened 2024-07-24
Convergence of FE trajectory with moving average


jaclark5:
It would be nice to see the moving average of the FE trajectory similar to the forward_backward_convergence plotting.

xiki-tempula:
merged

PR #381, opened 2024-07-24
Moving average

jaclark5:
Addition of Moving Average convergence and visualization functions, in addition to some changes to BAR to make this feasible. #380


jaclark5:
The convergence and visualization models don't appear to be present in the alchemtest respository. How should this be handled?

jaclark5:
I'm not sure how to resolve the outstanding codecov issues. After entering and executing the lines in an if-statement, codecov is labeling it as "partially covered"

xiki-tempula:
So codecov will label as partially covered if there isn't an explicit else statement and the else part has not been covered. I have not had a close look yet but in terms of metric I think the coverage of this PR is already decent.

jaclark5:
For reference, when this code was in the `lammps` branch, you asked why `states` was necessary instead of `self._states_`. That's because when running BAR with a single `fep-lambda` value (e.g., 0.1), while the number of `self._states_` reflects a larger number of states in the columns (e.g., 0.0, 0.1, 0.2, .. 1.0) then BAR complains about a mismatch between a matrix that is `(1,1)` [from fep-lambda] and one that is `(11,11)` [from `self._states_`].

xiki-tempula:
Do you mind give me a ping when you are happy for this PR to be reviewed? Thanks.

jaclark5:
@xiki-tempula Ok it's ready now!

docs/convergence.rst
```diff
@@ -75,6 +75,31 @@ is, where 0 fully-unequilibrated and 1.0 is fully-equilibrated. ::
...
+    >>> data_list = [extract_u_nk(xvg, T=300) for xvg in bz['Coulomb']]
+    >>> df = block_average(data_list, 'mbar')
+    >>> ax = plot_block_average(df)
+    >>> ax.figure.savefig('dF_t_block_average.png')
+
+Will give a plot looks like this
+
+.. figure:: images/dF_t_block_average.png
+
+   A convergence plot of showing that the forward and backward has converged
```
xiki-tempula, line 101: The description needs to be updated.

src/alchemlyb/convergence/convergence.py
```diff
@@ -94,7 +95,16 @@ def forward_backward_convergence(
         # select estimator class by name
         my_estimator = estimators_dispatch[estimator](**kwargs)
         logger.info(f"Use {estimator} estimator for convergence analysis.")
-
+    
+    # Check that each df in the list has only one value of lambda
```
xiki-tempula, line 99: Sorry do you mind remind me of why one cannot have more than one value of lambda?
I might be wrong but I think in principle, one could do forward_backward_convergence of more than one lambda?

src/alchemlyb/convergence/convergence.py
```diff
@@ -389,3 +399,122 @@ def A_c(series_list, precision=0.01, tol=2):
...
+    next 10% ... of the data.
+
+    Parameters
+    ----------
+    df_list : list
+        List of DataFrame of either dHdl or u_nk, where each represents a
+        different value of lambda.
+    estimator : {'MBAR', 'BAR', 'TI'}
+        Name of the estimators.
+        See the important note below on the use of "MBAR".
```
xiki-tempula, line 420: There is no important note, might be good to just remove it.

src/alchemlyb/convergence/convergence.py
```diff
@@ -389,3 +399,122 @@ def A_c(series_list, precision=0.01, tol=2):
...
+
+    if estimator in ["BAR"] and len(df_list) > 2:
+        raise ValueError(
+            "Restrict to two DataFrames, one with a fep-lambda value and one its forward adjacent state for a "
+            "meaningful result."
+        )
+        
+    # Choose length of comparison trajectory
+    lx_lambdas = [len(x) for x in df_list]
+    if len(set(lx_lambdas)) > 1:
```
xiki-tempula, line 475: I'm not quite sure if this shall be the default behaviour. I can see the rational for this. 
However, the data needs to be decorrelated before passing into this function so each lambda might have different number of data points and this behaviour might not be expected by everyone.
I think it would be better to have this as an option where one is to use the minimum and the other is just 1/10 of all the data points.

src/alchemlyb/convergence/convergence.py
```diff
@@ -389,3 +399,122 @@ def A_c(series_list, precision=0.01, tol=2):
...
+        logger.info("Moving Average Analysis: {:.2f}%".format(100 * i / num))
+        ind1, ind2 = lx // num * (i - 1), lx // num * i
+        sample = []
+        for data in df_list:
+            sample.append(data[ind1:ind2])
+        sample = concat(sample)
+        result = estimator_fit(sample)
+
+        average_list.append(result.delta_f_.iloc[0, -1])
+        if estimator.lower() == "bar":
```
xiki-tempula, line 497: Might be good to move this part to the BAR estimator as we kind of do this in several places.

src/alchemlyb/estimators/bar_.py
```diff
@@ -97,18 +97,21 @@ def fit(self, u_nk):
             (len(groups.get_group(i)) if i in groups.groups else 0)
             for i in u_nk.columns
         ]
-
+        states = [x for i, x in enumerate(self._states_) if N_k[i] > 0]
```
xiki-tempula, line 100: Would be good to add a test to show why this is needed. So we would always preserve this behaviour.

src/alchemlyb/estimators/bar_.py
```diff
@@ -97,18 +97,21 @@ def fit(self, u_nk):
...
         # Now get free energy differences and their uncertainties for each step
         deltas = np.array([])
         d_deltas = np.array([])
         for k in range(len(N_k) - 1):
+            if N_k[k] == 0 or N_k[k + 1] == 0:
```
xiki-tempula, line 105: Might be good to have a test for this as well.

src/alchemlyb/estimators/bar_.py
```diff
@@ -150,13 +153,11 @@ def fit(self, u_nk):
...
         # yield standard delta_f_ free energies between each state
-        self._delta_f_ = pd.DataFrame(
-            adelta - adelta.T, columns=self._states_, index=self._states_
-        )
+        self._delta_f_ = pd.DataFrame(adelta - adelta.T, columns=states, index=states)
 
         # yield standard deviation d_delta_f_ between each state
         self._d_delta_f_ = pd.DataFrame(
-            np.sqrt(ad_delta + ad_delta.T), columns=self._states_, index=self._states_
+            np.sqrt(ad_delta + ad_delta.T), columns=states, index=states
```
xiki-tempula, line 160: I think at this point, we shall overwrite the default behaviour of BAR.states to the new values.

src/alchemlyb/tests/test_convergence.py
```diff
@@ -26,6 +32,85 @@ def test_convergence_fep(gmx_benzene_Coulomb_u_nk, estimator):
     assert convergence.loc[9, "Backward"] == pytest.approx(3.04, 0.01)
 
 
+def test_block_average_ti(gmx_benzene_Coulomb_dHdl):
+    print(type(gmx_benzene_Coulomb_dHdl), type(gmx_benzene_Coulomb_dHdl[0]))
```
xiki-tempula, line 36: Don't use print in the tests

src/alchemlyb/tests/test_convergence.py
```diff
@@ -26,6 +32,85 @@ def test_convergence_fep(gmx_benzene_Coulomb_u_nk, estimator):
...
+
+@pytest.mark.parametrize("estimator", ["DUMMY"])
+def test_block_average_error_1(gmx_ABFE_complex_u_nk, estimator):
+    with pytest.raises(ValueError, match=r"Estimator DUMMY is not available .*"):
+        _ = block_average(gmx_ABFE_complex_u_nk, estimator)
+
+
+@pytest.mark.parametrize("estimator", ["MBAR"])
+def test_block_average_error_2_mbar(gmx_ABFE_complex_u_nk, estimator):
+    df_list = gmx_ABFE_complex_u_nk[10:15]
```
xiki-tempula, line 53: I kind of prefer that we would support this dataset.

src/alchemlyb/tests/test_visualisation.py
```diff
@@ -218,6 +218,47 @@ def test_plot_convergence_final_nan():
     plt.close(ax.figure)
 
 
+def test_plot_block_average(gmx_benzene_Coulomb_u_nk):
```
xiki-tempula, line 221: We might want to refactor this in to a class Instead of doing a couple test in a single function.

xiki-tempula (changes requested):
Thank you very much for the PR; it's very well-written! I have a couple of comments.

It would be important to consider supporting multiple lambda values. MD engines like Gromacs often utilize several lambda windows, with each lambda potentially changing at different rates. It could be problematic if this isn't supported.

src/alchemlyb/convergence/convergence.py
jaclark5: The purpose of this function is assess whether a particular production run has converged. The lambda state of that system must be constant throughout a dataframe for this assessment. If the lambda state changes later on in the trajectory (toward the bottom of the rows of the dataframe), the result of this function would not make sense or be useful. A user might eventually find their mistake, or they may think that their trajectory is not long enough. This check will help a user quickly produce useful results.

src/alchemlyb/convergence/convergence.py
jaclark5: The way alchemlyb has constructed the `forward_backward_convergence` function, a list of dataframes is passed to the function:
```
df_list = [
    df_1,
    def_2,
    ...
]
```
One choose a value for `num`, let's say 10. In line's 112 through 114 a fraction of the data is taken from each `df` and concatenated together:
```
        sample = []
        for data in df_list:
            sample.append(data[: len(data) // num * I])
```
So if `len(df_1) == 100` and `len(df_2) == 80`, then each free energy estimation will have 20% fewer states for the state represented in df_1 versus df_2.

I guess the real question is, when calculating free energy with MBAR or BAR, is it ok that the different states (w_r and w_f) are  different sizes? From the code in pymbar the answer seems to be no. I'll remove this, and update the documentation to make it more clear.

src/alchemlyb/estimators/bar_.py
jaclark5: I don't think so necessarily. self._states_ defines the number of lambda states in the columns, while states is capturing the number of states (for which the system is equilibrated to), so the number of unique fep-lambda values. So I guess the existing df._states_ this information is captures in df.columns, but I tried replacing it and there's a failure. I'm not willing to go through the code to reproduce that right now, but I think this is a larger question than should be answered in this PR.

src/alchemlyb/tests/test_convergence.py
jaclark5: The result of running that would not be meaningful since the lambda state changes over the course of a trajectory.

src/alchemlyb/tests/test_visualisation.py
jaclark5: I'm following the precedent of all the other tests in this file.

src/alchemlyb/convergence/convergence.py
xiki-tempula: Thank you for the clarification. I think there might be some misunderstanding here. It seems that you're referring to lambda dynamics, where a lambda value is constantly changing at different points in the trajectory. I agree that this isn't supported in this repository.

However, the use case I'm referring to involves MD engines like Gromacs, where multiple windows with different lambda values can run simultaneously. For example, you might have windows such as (coul lambda=0, vdw lambda=0), (coul lambda=0.5, vdw lambda=0), (coul lambda=1, vdw lambda=0), (coul lambda=1, vdw lambda=0.5), and (coul lambda=1, vdw lambda=1). Each lambda window represents an independent simulation, and within each simulation, the lambda value does not change.

The function in question would, for example, take the first 10% of data from all the windows to derive an MBAR estimate, then take the first 20% of data from all the windows to derive another MBAR estimate. This approach ensures that each independent lambda window is appropriately considered.

src/alchemlyb/convergence/convergence.py
xiki-tempula: Thank you for the detailed response, and I appreciate the clarification.

Regarding your question:

> I guess the real question is, when calculating free energy with MBAR or BAR, is it ok that the different states (w_r and w_f) are different sizes?

I think there might be some confusion here. While it's true that having different sample sizes for different lambda states can make some people uncomfortable, particularly because it gives more weight to states with more samples in the MBAR estimator, whether this is good or bad isn't entirely clear and might depend on the specific context.

I'm glad you've decided to remove the code in question, as it allows flexibility for users. If someone prefers that different lambda states have varying sample sizes, they can still input a list of dataframes with different lengths, and the function will handle them accordingly. On the other hand, if a user believes that each lambda state should have the same number of samples, they can input dataframes of equal length, ensuring that the function will provide an equal number of samples for each calculation.

My concern with the original code was that when the final point, which is supposed to use 100% of the data, ends up using a reduced dataset (due to equalizing the sample sizes across lambda states), it might yield different results compared to simply feeding 100% of the data into the estimator. This could potentially lead to discrepancies that users might not expect.

src/alchemlyb/convergence/convergence.py
jaclark5: I agree, and the way the `forward_backward_convergence` function handles that is to have each of those windows provided as a separate DataFrame in the df_list. This section is meant to ensure that each DataFrame has a constant set of lambda values independently, not that all the provided DataFrames contains the *same* set of lambda values.

src/alchemlyb/tests/test_convergence.py
jaclark5: Notice that if this test ran as `_ = block_average(df_list, estimator)` it would pass the way you want it to. I'm concatenating the DataFrames to make a set where the lambda state changes internally just to throw the error.

src/alchemlyb/convergence/convergence.py
jaclark5: Agreed!

src/alchemlyb/convergence/convergence.py
xiki-tempula: I see what you mean now. But I don't think the code is doing that? The `df.index` is the current lambda that is being simulated while `df.columns` is the set of all lambda values?

xiki-tempula: I see what you mean now. But what does 

`ind  = [j for j in range(len(lambda_values[0])) if len(list(set([x[j] for x in lambda_values]))) > 1][0]`

Do? I guess if you want the lambda value to be the same then 

```
if len(set([x[1:] for x in df.index.to_numpy()])) > 1:
    raise Exception
```
Should be enough?

jaclark5: I tried to make this flexible for the number of indices available in dataframe, as either fep-lambda, vdw-lambda, or coul-lambda could have multiple values. Given that `lambda_values` is a list of unique lambda sets, e.g., [[1]], [[0],[1]], [[0,0]], or [[0,0], [0,1]]. This line will identify the index that is changing so for [[0],[1]], ind=0, and for [[0,0], [0,1]], ind=1.

src/alchemlyb/convergence/convergence.py
xiki-tempula: I'm fine with change that.

jaclark5: I don't think this is something that should be changed in this PR. In the referenced issue comment Michael Shirts rejected a similar change because the standard deviations are correlated for BAR and a bootstrapping method is needed for accurate values, not a simple sqrt(sum(square())).

xiki-tempula: I agree. This could be in a different PR.

src/alchemlyb/estimators/bar_.py
jaclark5: I'm not sure how to test if this is needed or not so changed the way this is defined to more directly pull the lambda states in the indices and throw an error if those lambda states aren't represented in the columns of `u_nk`. This error is now tested with `test_block_average_error_3_bar`

src/alchemlyb/estimators/bar_.py
jaclark5: I'm not sure how to test this directly, so I added an error at the end of the loop so that if this if statement is true for every value of k, a descriptive error is given as to why. This error is tested with `test_block_average_error_4_bar`.

xiki-tempula (approved):

orbeckst:
@jaclark5 we forgot to add you to AUTHORS in this PR. 

I'll fix it in PR #395 .

PR #397, opened 2024-09-16
Resolve bug in bar_model.delta_f_ creation

jaclark5:
In the recent PR #381, there was a change to creating the delta_f_ matrix creation which resulted in the columns and indices being tuples that were in the wrong order for single lambda computations. 

This PR resolves that issue, ensures that columns are in the correct order, and adds a test for the delta_f_ columns in the future.


orbeckst (approved):
lgtm & thanks for explicitly testing!

PR #399, opened 2024-09-17
[doc] tutorial: use alchemlyb.concat

orbeckst:
replace `pandas.concat()` with `alchemlyb.concat()` in the tutorial (given that we explicitly tell users to use it)


orbeckst:
@xiki-tempula we could add this fix with the 2.4.1 release #403

xiki-tempula (approved):

PR #406, opened 2024-09-19
Add entropy and enthalpy

jaclark5:
Add enthalpy and entropy outputs for MBAR and BAR (via 2-state MBAR).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,9 +18,7 @@ The rules for this file:
 
 Enhancements
   - Parallelise read and preprocess for ABFE workflow. (PR #371)
-  - Added matrices for entropy and enthalpy for MBAR and BAR,
-    when BAR is substituted with a series of 2-state MBAR calculations.
-    (PR #406)
+  - Added matrices for entropy and enthalpy for MBAR (PR #406)
 
 
 09/19/2024 orbeckst, jaclark5
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,14 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
   * release numbers follow "Semantic Versioning" https://semver.org
 
+??/??/2024 jaclark5
+
+  * 2.4.1
+
+Enhancements
+ - Added matrices for entropy and enthalpy for MBAR and BAR,
+   when BAR is substituted with a series of 2-state MBAR calculations.
+   (PR #406)
 
 09/17/2024 jaclark5, orbeckst
 
--- a/CHANGES
+++ b/CHANGES
@@ -17,8 +17,8 @@ The rules for this file:
   * 2.5.0
 
 Enhancements
-  - Parallelise read and preprocess for ABFE workflow. (PR #371)
   - Added matrices for entropy and enthalpy for MBAR (PR #406)
+  - Parallelise read and preprocess for ABFE workflow. (PR #371)
 
 
 09/19/2024 orbeckst, jaclark5
--- a/src/alchemlyb/estimators/bar_.py
+++ b/src/alchemlyb/estimators/bar_.py
@@ -185,8 +185,16 @@ class BAR(BaseEstimator, _EstimatorMixOut):
             if use_mbar:  # now determine df and ddf using pymbar.MBAR
                 tmp_u_nk = concat(
                     [
-                        groups.get_group(self._states_[k]),
-                        groups.get_group(self._states_[k + 1]),
+                        groups.get_group(
+                            self._states_[k]
+                            if isinstance(self._states_[k], tuple)
+                            else (self._states_[k],)
+                        ),
+                            groups.get_group(
+                                self._states_[k + 1]
+                                if isinstance(self._states_[k + 1], tuple)
+                                else (self._states_[k + 1],)
+                        ),
                     ]
                 )
                 columns = tmp_u_nk.columns.tolist()
--- a/src/alchemlyb/estimators/bar_.py
+++ b/src/alchemlyb/estimators/bar_.py
@@ -1,10 +1,8 @@
 import numpy as np
 import pandas as pd
-import pymbar
 from pymbar.other_estimators import bar as BAR_
 from sklearn.base import BaseEstimator
 
-from .. import concat
 from .base import _EstimatorMixOut
 
 
@@ -20,16 +18,9 @@ class BAR(BaseEstimator, _EstimatorMixOut):
     relative_tolerance : float, optional
         Set to determine the relative tolerance convergence criteria.
 
-    method : str, optional, default='false-position' or 'robust'
-        Choice of method to solve BAR nonlinear equations,
-        one of 'self-consistent-iteration' or 'false-position' (default: 'false-position').
-        If ``use_mbar == True`` the default is 'robust', see 
-        :class:`~alchemlyb.estimators.mbar_.MBAR` for options.
-
-    n_bootstraps : int, optional
-        Whether to use bootstrap to estimate uncertainty. `0` means use analytic error
-        estimation. 50~200 is a reasonable range to do bootstrap. Currently used when
-        ``use_mbar is True``.
+    method : str, optional, default='false-position'
+        choice of method to solve BAR nonlinear equations,
+        one of 'self-consistent-iteration' or 'false-position' (default: 'false-position')
 
     verbose : bool, optional
         Set to True if verbose debug output is desired.
@@ -43,20 +34,6 @@ class BAR(BaseEstimator, _EstimatorMixOut):
     d_delta_f_ : DataFrame
         The estimated statistical uncertainty (one standard deviation) in
         dimensionless free energy differences.
-
-    delta_h_ : DataFrame
-        The estimated dimensionless enthalpy difference between each state.
-
-    d_delta_h_ : DataFrame
-        The estimated statistical uncertainty (one standard deviation) in
-        dimensionless enthalpy differences.
-
-    delta_s_ : DataFrame, optional
-        The estimated dimensionless entropy difference between each state.
-
-    d_delta_s_ : DataFrame
-        The estimated statistical uncertainty (one standard deviation) in
-        dimensionless entropy differences.
 
     states_ : list
         Lambda states for which free energy differences were obtained.
@@ -80,7 +57,6 @@ class BAR(BaseEstimator, _EstimatorMixOut):
     .. versionchanged:: 2.4.0
        Added assessment of lambda states represented in the indices of u_nk
        to provide meaningful errors to ensure proper use.
-       Added computation of enthalpy and entropy with 2-state MBAR
 
     """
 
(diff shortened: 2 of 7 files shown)


src/alchemlyb/estimators/bar_.py
```diff
@@ -65,29 +89,47 @@ def __init__(
...
 
+        Sets the attributes: delta_f_ and d_delta_f_
+
         Parameters
         ----------
         u_nk : DataFrame
             u_nk[n,k] is the reduced potential energy of uncorrelated
             configuration n evaluated at state k.
 
+        use_mbar : bool, optional, default=False
```
xiki-tempula, line 118: Sorry do you mind explain to me why do you want to do this in BAR? I think if you want the enthalpic and entropic contributions, you would just run MBAR?

mrshirts:
@jaclark5 and I have been talking over things (she can discuss her side here) and the current thought is that MBAR is better for entropy and enthalpies, though there are some questions about the calculations she is still interested in. 

The BAR algorithm is more surely converging since there are better methods for 1-dimensional root finding (I think I coded up false position) than N-dimensional root finding.  One can start with BAR, and then use those solutions to initialize 2-state MBAR (should just take 1-2 steps as it locks in numerically, as the values should be the same).

jaclark5:
So the suggestion is to remove the `use_mbar` flag, and instead run 2-state mbar only for the entropies / enthalpies? If that's the case, then if `calculate_enthalpy_entropy is True` then BAR.fit() will run BAR and 2-state MBAR  in `fit()`. I can make that change if it sounds agreeable.

xiki-tempula:
Would this kind of interface work?
```
bar = BAR()
bar.fit(u_nk)
mbar=MBAR(initial_guess=bar._delta_f) # Need to do some data massage to be able to get this in as initial guess.
mbar.fit(u_nk, compute_entropy_enthalpy=True)
```
Note that if you just need to use BAR to provide an initial guess for the MBAR, you could just do
```
mbar=MBAR(initial_guess='bar') 
mbar.fit(u_nk, compute_entropy_enthalpy=True)
```

mrshirts:
> I think... I would want to test how MBAR handles a sparse matrix. 

It can't, it will give incorrect results.  It's not equivalent.

The advantage of a BAR interface to enthalpy-entropy, I see, is when you did NOT collect the full MBAR data, and only have neighboring values.

> Although it seems that the enthalpy and entropy values for the gromacs benzene example are pretty different for full MBAR vs 2-state MBAR case so I'

Yeah, I realize this is still pending with me to look at, things have been piling up.

jaclark5:
Ok, so if there is merit to offering a calculation of 2-state MBAR (which I have need for, and it seems could be more generally used as-well), there are a few options:
1) Keep the PR as is where the BAR class can contain a 2-state MBAR option within the existing for-loop structure for BAR
2) Leave the BAR class alone, and and add a 2-state MBAR option to the MBAR class and add the same sort of for-loop structure that is present in the BAR class.
3) *Fill in the blank*

xiki-tempula:
For the MBAR part, I think I'm happy with the current state, where a simple call to pymbar gives the entropy and enthalpy.
With regard to the BAR part, I think it is a rather peculiar use case. We might need more discussion for it to be included.

jaclark5:
Hey @xiki-tempula, I removed the changes to BAR. I was able to resolve my issues with MBAR and LAMMPS. I think you'll find this PR acceptable.

src/alchemlyb/estimators/mbar_.py
```diff
@@ -57,6 +57,20 @@ class MBAR(BaseEstimator, _EstimatorMixOut):
         The estimated statistical uncertainty (one standard deviation) in
         dimensionless free energy differences.
 
+    delta_h_ : DataFrame
+        The estimated dimensionless enthalpy difference between each state.
```
xiki-tempula, line 61: Is enthalpy dimensionless? Should it be having energy unit?

CHANGES
```diff
@@ -12,12 +12,13 @@ The rules for this file:
...
   * release numbers follow "Semantic Versioning" https://semver.org
 
-**/**/**** xiki-tempula
+**/**/**** xiki-tempula, jaclark5
 
   * 2.5.0
 
 Enhancements
   - Parallelise read and preprocess for ABFE workflow. (PR #371)
+  - Added matrices for entropy and enthalpy for MBAR (PR #406)
```
xiki-tempula, line 21: The latest goes to the top.

src/alchemlyb/estimators/mbar_.py
```diff
@@ -187,9 +210,27 @@ def fit(self, u_nk):
...
+            self._delta_s_ = pd.DataFrame(
+                out["Delta_s"], columns=self._states_, index=self._states_
+            )
+            self._d_delta_s_ = pd.DataFrame(
+                out["dDelta_s"], columns=self._states_, index=self._states_
+            )
 
         self._delta_f_.attrs = u_nk.attrs
         self._d_delta_f_.attrs = u_nk.attrs
+        if compute_entropy_enthalpy:
```
xiki-tempula, line 229: I see that you do add the units back.

src/alchemlyb/estimators/mbar_.py
jaclark5: Do you want me to change the docstring? I put this since the existing documentation for free energy is:
```
    delta_f_ : DataFrame
        The estimated dimensionless free energy difference between each state.
```
and I'm treating enthalpy and entropy the same as free energy... it seems I should change entropy appropriately though.

xiki-tempula: Ok, then I think this is fine. I will change this in a later PR.

src/alchemlyb/estimators/mbar_.py
```diff
@@ -187,9 +211,29 @@ def fit(self, u_nk):
...
+            self._delta_h_.attrs = u_nk.attrs
+            self._d_delta_h_.attrs = u_nk.attrs
+            self._delta_s_.attrs = u_nk.attrs
+            self._d_delta_s_.attrs = u_nk.attrs
+            self._delta_s_.attrs["energy_unit"] = "k"
```
xiki-tempula, line 235: Why does entropy has unit of k? I think it shall be `u_nk.attrs["energy_unit"] + '/K'`?

xiki-tempula: Personally, I'm fine for this unit to be the same as other energy unit as it would mean that the unit conversion would work as it is (between J and cal).

jaclark5: Free energy is in units of energy (J) and when scaled by kT, is in units of kT. Similarly entropy is in units of energy/unit temperature (J/K) and so is scaled by k, with the units of k.

mrshirts: Reduced entropy should be unitless, entropy should be J/degree, right?  Even if passed through some unitless quantities, probably the code should be clear what is going on.

jaclark5: I think it's important to have the correct units, so I'll look into the unit conversion, I didn't consider that feature.

mrshirts: I think we need to be a little careful when talking about things being in units of KT, this basically means unitless - all reduced quantities are unitless.  I think being in the units of KT, are "unitless here, I'm keeping track of if we were to convert them back to our physical system, what the units would be", though I'm rushing to get kids of to school so can't to into it carefully here with respect to the actual code.

jaclark5: The way alchemlyb defines the units as `u_nk.attrs['energy_unit'] = "kT"` I think implies that it's dimensionless while also indicating what quantities should be used to restore the units (although I hope that an average user would know). This is also why I would want the entropy to similarly have the correct scaling quantity, 'k'.

xiki-tempula: The `energy_unit`, allowed to be kT; kcal/mol and kJ/mol, it is just defaults to kT. I think `k` is more like a scientific constant for most people so having it as unit is kind of bad.
The unit conversion is supported between these three units, which is just a scaling factor.
So the easiest solution is to just set the entropy to have the same unit as enthalpy.

jaclark5: To adjust the unit conversion functionality to accommodate entropy, should I add the appropriate duplicated function of `to_kT` as `to_k` or add if-statements to detect the case where it's entropy? I'm leaning toward the former since running a matrix of entropy values through `to_kT()` would feel wrong. This would mean I would duplicate `to_kcalmol(df, T=None)` as `to_kcalmolK(df)` and `to_kJmol(df, T=None)` as `to_kJmolK(df)`.

src/alchemlyb/estimators/mbar_.py
```diff
@@ -187,9 +211,27 @@ def fit(self, u_nk):
...
             out["Theta"], columns=self._states_, index=self._states_
         )
+        if compute_entropy_enthalpy:
+            self._delta_h_ = pd.DataFrame(
+                out["Delta_u"], columns=self._states_, index=self._states_
+            )
+            self._d_delta_h_ = pd.DataFrame(
+                out["dDelta_u"], columns=self._states_, index=self._states_
+            )
+            self._delta_sT_ = pd.DataFrame(
```
xiki-tempula, line 221: I'm happy for this value to be s*T. Is the mbar output s or s*T?

src/alchemlyb/parsing/__init__.py
```diff
@@ -31,8 +31,13 @@ def wrapper(outfile, T, *args, **kwargs):
         dict_with_df = func(outfile, T, *args, **kwargs)
         for k in dict_with_df.keys():
             if dict_with_df[k] is not None:
+
+                if T < 273:
```
xiki-tempula, line 35: Would it be better for this to be a warning?
I mean if T is a negative number then it is definitely an error but if it is just below 0 C then it is not strictly impossible.

src/alchemlyb/estimators/mbar_.py
jaclark5: Well the mbar output is reduced entropy, so S/k or s*T/(k*T) so it's consistent.

src/alchemlyb/parsing/__init__.py
jaclark5: I can remove this since I took the liberty of adding it, but since you have a value of the gas constant that is hard coded to be in Kelvin and the docs are defined in Kelvin, this will always be true and catch the error that someone accidentally entered a value in C.

src/alchemlyb/estimators/mbar_.py
xiki-tempula: I know that the unit is the same but I worried about the magnitude.
The G=H-sT so if the `Delta_s` is s then the `Delta_f` should be the same as `Delta_u`-`Delta_s` * T. If Delta_s is sT, then `Delta_f` should be `Delta_u`-`Delta_s`

jaclark5: Yes I agree, so I'm changing the variable names to make that clear.  I'll do whatever the maintainers agree on, maybe @mrshirts or @orbeckst have an opinion?

**Summary**: Entropy has different units and which need to be handled properly. While the units of enthalpy and free energy are `attr['energy_units'] = 'kT', 'kcal/mol' or 'kJ/mol'`, the equivalent for entropy would then be `'k', 'kcal/mol/K' or 'kJ/mol/K'`. The downside of this difference in units would require either having copies of the units.py functions for entropy specifically or add a series of if statements to the functions in units.py to detect and handle entropy units. Also, it's awkward to say that dimensionless entropy has units of `'k'`

**Alternative**: By having the entropy matrix represent S*T (denoted with the variable name `delta_sT_`) the object can retain the same units as enthalpy and free energy and use the same unit conversion functions, making everything cleaner internally, and avoid the awkward unit "definition" of `'k'` for S.

xiki-tempula (approved):
LGTM

src/alchemlyb/estimators/mbar_.py
xiki-tempula: Thanks. I'm merging this PR. I have checked the numerical values and can verify that Delta_s is not s but rather s*T. So it is in the unit of kT, kJ/mol and kcal/mol.

Issue #421, opened 2024-10-16
Forward and backward convergence visualization example in docs does not work


spackman:
The visualization example for forward and backward convergence in the docs fails on the line: 
   
`ax = plot_convergence(forward, forward_error, backward, backward_error)`  
  
This is because each argument passed to plot_convergence() is a list, and the function expects a dataframe with labeled columns.  
  
See the link to the docs here:
https://alchemlyb.readthedocs.io/en/latest/visualisation.html

PR #422, opened 2024-10-16
correct error in tutorial for plot_convergence()

spackman:
The function plot_convergence() expects a pandas dataframe as the single argument in the current implementation, causing it to fail when passed 4 different lists as arguments instead. This edit corrects the tutorial to appropriately pass a single dataframe with the appropriate column names, resulting in code that successfully reproduces the example graph.  
  
This should resolve issue #421.


xiki-tempula (approved):
LGTM Thanks for spotting it

PR #429, opened 2025-05-12
Use bar_intervall instead of ntpr for parsing AMBER MBAR timestamps

### LINKED ISSUE #428: Incorrect timestamp parsing in AMBER output
## Expected Behavior
The timestamp for each MBAR analysis point should be accurate when parsing AMBER output.

## Current Behavior
If `ntpr` (the energy print frequency) and `bar_interval` (the interval for MBAR analysis) differ, the parser incorrectly assumes `ntpr` as the sampling interval, leading to an inaccurate time series.

yuxuanzhuang:
Fix #428


xiki-tempula:
Thanks for the PR, do you mind

- Add a new test to check the amber parser is returning the correct time.
- Modify the https://github.com/alchemistry/alchemlyb/blob/master/CHANGES
- Add yourself to https://github.com/alchemistry/alchemlyb/blob/master/AUTHORS

yuxuanzhuang:
I assume the test file needs to be put in https://github.com/alchemistry/alchemtest.

Could you help me with it? see the output attached!

[timestamp.txt](https://github.com/user-attachments/files/20277717/timestamp.txt)

src/alchemlyb/tests/parsing/test_amber.py
```diff
@@ -282,3 +282,13 @@ def test_concatenated_amber_dhdl(tmp_path):
         match="TI Energy detected after the TIMINGS section.",
     ):
         extract(tmp_path / "amber.out", 298)
+
+
+def test_correct_timestamp_u_nk(testfiles):
```
xiki-tempula, line 287: @yuxuanzhuang Thanks for adding the test. I think it is fine to just adapt this test to use an existing file and check if the timestep is correct would be enough. No need to add a new file.

AUTHORS
```diff
@@ -59,3 +59,6 @@ Chronological list of authors
 
 2024
   - Jennifer A. Clark (@jaclark5)
+
+2025
+  - Yuxuan Zhuang (@yuxuanzhuang)
```
xiki-tempula, line 64: blank line after this

src/alchemlyb/tests/parsing/test_amber.py
yuxuanzhuang: thanks. then I think `test_u_nk_time_reading(single_u_nk)` is already checking it.

xiki-tempula: Ok, so it was already checked, I was thinking something along the line of 
```
def test_u_nk_time_reading_bar_intervall(single_u_nk, tmp_path):
    """Test if time information is read correctly when extracting u_nk"""
    with bz2.open(single_u_nk, 'rt') as bz2_file:
        content = bz2_file.read()

    with open(tmp_path / 'amber.log', 'w') as text_file:
        text_file.write(content.replace("bar_intervall =     1000", "bar_intervall =     100"))
    u_nk = extract_u_nk(tmp_path / 'amber.log', T=298.0)
    assert_allclose(u_nk.index.values[0][0], ??)
```
Then you show that changing the bar_intervall in the log file will change the time stamp in the generated dataframe.

xiki-tempula:
Thanks. The test looks good. Do you mind fix the formatting and this PR is ready to be merged?

xiki-tempula:
Yeah, you are right. I will fix the test and merge your PR.

xiki-tempula (approved):
LGTM. Thanks!

PR #430, opened 2025-05-15
Fix GMX test

xiki-tempula:
I have no idea why GMX failed, I have downgraded scipy, numpy and pandas to version before that but still cannot reproduce the old result. Consider that the difference is small, I think we shall just correct the test.


orbeckst (approved):
Sounds sensible. LGTM

PR #431, opened 2025-06-12
Fix test

xiki-tempula:


pyproject.toml
```diff
@@ -31,18 +31,17 @@ classifiers = [
...
     "Topic :: Scientific/Engineering :: Bio-Informatics",
     "Topic :: Scientific/Engineering :: Chemistry",
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 
 dependencies = [
-	     "numpy",
+	     "numpy<2.3",
```
orbeckst, line 44: If we only support numpy up to a specific version then we need to state it together with a reason in the CHANGELOG.

orbeckst (changes requested):
Interesting that you narrowed it down to numpy 2.3.

Do we know if this is in pymbar or in alchemlyb itself?

I'm ok for right now to say that alchemlyb only works as expected with np <2.3 but we need to **raise an issue for "numpy 2.3 gives different results"** and look into what exactly is happening and figure out how big the differences are. Without seeing more, I wouldn't want to say "it's fine, we just get different results" — having these tests is exactly what should guard against random changes in the super-complicated Python ecosystem.

What's your opinion?

pyproject.toml
orbeckst: The numpy issue shows as an example corner cases with nan. We are seeing issues in normal calculations. 

Do we know if this is ultimately pymbar or alchemlyb?

Could you please post (in the main comments) the output when our tests run with numpy 2.3 so that I can get a sense of differences? At the face of it, I don't want my free energies to differ by 10% (rtol=0.1) just because I switched to a new version.

xiki-tempula: 100% pymbar madness https://github.com/choderalab/pymbar/issues/556

xiki-tempula:
Do we want get the test fixed first and have the PR https://github.com/alchemistry/alchemlyb/pull/429 merged?

pyproject.toml
orbeckst: Thanks for checking and raising the issue in pymbar!

orbeckst: Yikes for the differences. 😱

orbeckst (changes requested):
Really good that you found that switching from numpy <2.3 to numpy 2.3 makes pymbar give changes in output by 0.1 kT (https://github.com/choderalab/pymbar/issues/556).

I would feel much safer if we **restrict to numpy <2.3** for right now. I'd
- add a comment to the tests and reference this PR and the pymbar PR
- add an entry to CHANGES under *Changes*: For the time being, only numpy <2.3 is supported due to potential changes in output of up to 0.1 kT (see PR #431 and issue https://github.com/choderalab/pymbar/issues/556 )

Would you be ok with this conservative approach?

PR #432, opened 2025-06-27
Using ruff to do format and lint check

xiki-tempula:


src/alchemlyb/parsing/amber.py
```diff
@@ -223,15 +223,13 @@ def file_validation(outfile):
...
 
-        mbar_ndata = 0
         have_mbar, bar_intervall, mbar_states = secp.extract_section(
             "^FEP MBAR options:",
             "^$",
             ["ifmbar", "bar_intervall", "mbar_states"],
             "^---",
         )
         if have_mbar:
-            mbar_ndata = int(nstlim / bar_intervall)
```
orbeckst, line 234: I assume removing these was fine and you did that on purpose. tests are passing so ... lgtm

src/alchemlyb/parsing/gmx.py
```diff
@@ -118,21 +116,21 @@ def extract_u_nk(xvg, T, filter=True):
...
                 for t in thermo_state:
                     v.append(statevec[int(t)][i])
-                u_k[l] = v
+                u_k[lambda_value] = v
         else:
             state_legend = _extract_legend(xvg)
-            for i, l in enumerate(state_legend):
-                u_k[l] = state_legend[l]
+            for i, state_legend_item in enumerate(state_legend):
+                u_k[state_legend_item] = state_legend[state_legend_item]
```
orbeckst, line 127: that's probably more verbose than necessary but ok

src/alchemlyb/visualisation/ti_dhdl.py
```diff
@@ -118,7 +118,7 @@ def plot_ti_dhdl(dhdl_data, labels=None, colors=None, units=None, ax=None):
             )
 
     # Get the real data out
-    xs, ndx, dx = [0], 0, 0.001
+    xs, ndx, _dx = [0], 0, 0.001
```
orbeckst, line 121: Why do we assign dx if we don't use it... Did you check that `_dx` is ok??

src/alchemlyb/parsing/amber.py
xiki-tempula: Yes, I had a look at this and it does seems that mbar_ndata is not used.
The reason that I want to add linting check is to prevent this from being merged in the first place.

PR #433, opened 2025-06-27
Use mypy to do type check

xiki-tempula:


xiki-tempula:
@orbeckst Do you feel maybe full typing support is too much?

orbeckst:
Sorry @xiki-tempula , this PR just slipped of my radar. My silence does not mean that it's a bad idea. In fact, typing is a good idea – although I find the code to look noisier, it makes it more robust, and ultimately that's important for maintainable scientific code. So 👍 from me.

orbeckst (approved):
Tests pass and this is a good idea.

PR #441, opened 2025-10-21
update type hinting

orbeckst:
Local type checker
```
mypy src alchemlyb
```
is happy again.


src/alchemlyb/estimators/mbar_.py
```diff
@@ -172,10 +172,10 @@ def fit(self, u_nk: pd.DataFrame, compute_entropy_enthalpy: bool = False) -> "MB
                 ]
                 initial_f_k = pd.Series(
                     [
-                        initial_f_k.loc(x) if x in states else np.nan
```
orbeckst: Not really sure how `loc(x)` worked before ...

